PR #12, opened 2023-05-08
Initial pull request for lensed SNe

jpierel14:


requirements.txt
```diff
@@ -0,0 +1,5 @@
+numpy
+sncosmo
```
sibirrer, line 2: can you specify the minimum version number to make sure that the environment is set up well even if people have old version of sncosmo?

sim_pipeline/__init__.py
```diff
@@ -2,3 +2,6 @@
 
 __author__ = """DESC/SLSC"""
 __version__ = '0.1.0'
+
+from .transient_generation import *
```
sibirrer, line 6: just wondering whether we need these generic inputs or not.

sim_pipeline/sim_pipeline.py
```diff
@@ -1 +1,23 @@
 """Main module."""
+
+import numpy as np
+import copy
+import matplotlib.pyplot as plt
+
+from .transient_generation import LensedTransientImage,LensedTransientSystem
+
+
+def LensedSystem_from_LensParams(lensed_transients, time_delays, magnifications):
```
sibirrer, line 10: can you document the inputs to this definition? Also, definitions are lower_case_wordings

tests/test_sim_pipeline.py
```diff
@@ -2,23 +2,55 @@
 
 """Tests for `sim_pipeline` package."""
 
-import pytest
+import unittest
```
sibirrer, line 5: can you write the test functions in files and folders that mimic the sim-pipeline file structure such that we all know where exactly to look for a test function for a specific definition

sibirrer (changes requested):
Thanks a lot @jpierel14! I just made some minor comments/request mainly on the structure such that merging and co-development works well. Have not reviewed the code itself. Also let me know if you encounter issues with the testing. Seems that the tests fail for some reason.

requirements.txt
jpierel14: Updated

sim_pipeline/sim_pipeline.py
jpierel14: This function is sort of temporary, pending the connection point to the lens sim part so I'll leave it for now

tests/test_sim_pipeline.py
jpierel14: Updated

PR #18, opened 2023-06-22
updated galaxy ellipticity input function

sibirrer:
minor change to hopefully mitigates errors reported by Narayan Khadka


sibirrer:
@nkhadka21 thanks for reporting an error that you got. I mildly changed the function here. Let me know if things do not work in a Github issue to keep track of it

PR #24, opened 2023-07-21
removed my home path from gg-lensing tutorial.

nkhadka21:
Hi Simon, this is just a test pull request! I have only removed my RSP home path from the tutorial.


sibirrer:
This looks good! In the future, I would suggest removing the generated plots from the notebook such that the file size is less (just re-set the notebook). Currently the notebook is 1.5Mb and every time one makes a new plot and commits it, it adds to the size of this repo

PR #48, opened 2023-09-06
fix most of the docstring errors

sibirrer:
I made some minor changes to hopefully fix some docstrings in readthedocs


sibirrer:
@ajshajib I expected that the pre-commit is formatting the lines to fulfill the PEP8 standards with black, or I am missing something?

ajshajib:
@sibirrer, that's what it's expected to do, yes. But, it doesn't always work perfectly, unfortunately, particularly `docformatter` for docstring formatting. The `black` formatter itself works very well in my experience, but it only works on the code lines, not the docstrings. So, we still have to keep an eye out for blemishes. One reason the `docformatter` could break, is when it sees an unexpected way the docstring is laid out.  For example, the expected Sphinx option is `:return:` and not `:returns:`.

ajshajib (approved):

PR #50, opened 2023-09-06
Gg lensing

nkhadka21:
Just removed galsim dependence from image_simulation.py.

### DIFF
--- a/sim_pipeline/image_simulation.py
+++ b/sim_pipeline/image_simulation.py
@@ -1,5 +1,4 @@
 from lenstronomy.SimulationAPI.sim_api import SimAPI
-import galsim
 from astropy.visualization import make_lupton_rgb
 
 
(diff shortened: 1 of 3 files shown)


sibirrer:
@nkhadka21 there are some conflicts to be solved. Best is to solve them on your end

nkhadka21:
> @nkhadka21 there are some conflicts to be solved. Best is to solve them on your end

Resolved!

sim_pipeline/image_simulation.py
```diff
@@ -100,32 +99,6 @@ def sharp_image(
     return image
 
 
-def galsimobj(image, pix_scale, flux):
```
sibirrer, line 103: you removed this definition but did not change any other code. Was this definition unused or perhaps used in a jupyter notebook?

sibirrer (commented):
I have a comment about one definition that you removed that was presumably unused

sim_pipeline/image_simulation.py
nkhadka21: This was used in jupyter notebook. I removed everything related to this.

sibirrer (approved):
Looks good!

PR #52, opened 2023-09-07
added time delays and observer times for time series predictions

sibirrer:


sibirrer:
Hi @nkhadka21, I made two routines for the time delay prediction. Make sure you understand those or otherwise let me know

nkhadka21 (approved):

PR #58, opened 2023-09-12
Add gg_lens_table functionality and tests

Newton-Lightspeed:
This pull request introduces two new files to enhance the sim-pipeline:

gg_lens_table.py:

This module provides functionality to generate an astropy table for a given gg_lens population.
It collects various lens properties such as Einstein Radius, Deflector Ellipticity, Deflector Velocity Dispersion, and more.
This module ensures efficient extraction of properties using a single loop and lets the user select which properties to extract.
gg_lens_table_test.py:

This test file ensures the functionality and serves as a test case only focusing on extracting one parameter. of the gg_lens_table.py.

Additional Comments:

The code has been refactored based on peer review.
Requesting a review to merge these enhancements into the main branch. Feedback and suggestions are most welcome!

### DIFF
--- a/sim_pipeline/gg_lens_table.py
+++ b/sim_pipeline/gg_lens_table.py
@@ -1,0 +1,55 @@
+from astropy.table import Table
+from sim_pipeline.Pipelines.skypy_pipeline import SkyPyPipeline
+from sim_pipeline.gg_lens import GGLens, theta_e_when_source_infinity
+import numpy as np
+
+
+
+def create_table(gg_lens_population, parameters=['einstein_radius', 'ellipticity', 'velocity_dispersion', 'stellar_mass', 'magnitude', 'source_redshift', 'lens_redshift']):
+    """
+    Creates an astropy table for a given gg_lens population.
+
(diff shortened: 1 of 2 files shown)


sim_pipeline/gg_lens_table.py
```diff
@@ -0,0 +1,65 @@
...
+        "magnitude",
+        "source_redshift",
+        "lens_redshift",
+    ],
+):
+    """Creates an astropy table for a given gg_lens population.
+
+    :param gg_lens_population: List of GGLens instances.
+    :type gg_lens_population: list
+    :param parameters: List of parameters to extract from the population.
```
sibirrer, line 23: this deserves a bit more explanation. These are de-facto definitions in the gg_lens class

sibirrer (changes requested):
thanks @Newton-Lightspeed. I like your re-structuring of the method. The tests fail and you should move the test definitions into the tests/ folder and remove all the code from files outside of definitions, otherwise these get executed when loading the file

PR #62, opened 2023-09-13
solve issue with point sources and simplifies image_position() calculation

sibirrer:


sibirrer:
@nkhadka21 should be good now.

nkhadka21 (approved):

PR #71, opened 2023-09-27
implemented velocity dispersion function

sibirrer:


sibirrer:
@ajshajib it seems like, again, that the black formatter is not pushing the format changes. Did this just change again?

sibirrer:
@nkhadka21 do you happen to know why the pre-commit PR does not work?

sim_pipeline/Deflectors/velocity_dispersion.py
```diff
@@ -11,6 +11,61 @@
...
+    a NFW halo profile, assuming isotropic anisotropy
+
+    :param r: radius of the luminosity-weighted velocity dispersion [arcsec]
+    :param m_star: stellar mass [M_sun]
+    :param rs_star: stellar half light radius [physical kpc]
+    :param m_halo: Halo mass [physical M_sun]
+    :param c_halo: halo concentration
+    :return: velocity dispersion [km/s]
+    :param cosmo: cosmology
+    :type cosmo: ~astropy.cosmology class
```
ajshajib, line 26:
```suggestion
    :param cosmo: cosmology
    :type cosmo: ~astropy.cosmology class
    :return: velocity dispersion [km/s]
```

sibirrer:
@ajshajib it still fails, probably due to different black versions. Also, how do I run these locally? It really should not happen and I wonder whether black adds more troubles than it solves

ajshajib:
I don't exactly know why this is failing. Currently, it's failing due to `docformatter`, which is probably the main cause for the previous failures as well.

To run locally, just `pip install black` and `pip install docformatter`, then do:

```
black .
docformatter ./* -r --black --in-place
```

ajshajib:
I only ran `black` in my last commit, so please try running `docformatter` on your local machine to check if that solves the problem for now.

ajshajib:
Pre-commit CI gives this message:

> pre-commit.ci was unable to push due to branch protection rules

Did anything change in the branch protection rules recently?

sibirrer:
> Pre-commit CI gives this message:
> 
> > pre-commit.ci was unable to push due to branch protection rules
> 
> Did anything change in the branch protection rules recently?

hmm, this could have been the issue on my side

sibirrer:
Thank you @ajshajib! I could manually solve it and added the commands in the make file as well in this PR, should be good to go once the tests pass

ajshajib:
A point on `pre-commit` is that the originally intended workflow for `pre-commit` is to have it run locally. The name `pre-commit` comes from the fact that if it is installed locally, then the hooks (for `black`, `docformatter`, etc.) would automatically run locally at every `git commit` command ("pre" of every attempted commit). The way I added the pre-commit system to this repo is a workaround as more like a "post-commit" enforced at the main repo level on GitHub.

Preferably, `pre-commit` needs to be installed by every developer with this:

```
pip install pre-commit
cd /path/to/sim-pipeline/directory
pre-commit install
```

In that case, `pre-commit` should be added to the requirements.txt.

PR #73, opened 2023-09-29
Add cosmology support in the galaxies pipeline

trivialTZ:
**Add cosmology support in the galaxies pipeline**

- Enhanced pipeline configuration to support various cosmology models.
- Ensured compatibility with the existing pipeline structure and functionality.
- Enhanced the `GalaxyGalaxyLensPop` class to accept dynamic cosmology inputs.
- Fixed issues ensuring the cosmological input directly influences galaxy generation.
@sibirrer


sim_pipeline/Pipelines/skypy_pipeline.py
```diff
@@ -35,6 +35,27 @@ def __init__(self, skypy_config=None, sky_area=None, filters=None):
                 new_fsky = f"fsky: {sky_area.value} {sky_area.unit}"
                 content = content.replace(old_fsky, new_fsky)
 
+            if cosmo:
```
sibirrer, line 38: I would specify
if cosmo is not None:

tests/test_Pipelines/test_skypy_pipeline.py
```diff
@@ -8,6 +9,30 @@ def setup_method(self):
         sky_area = Quantity(value=0.05, unit="deg2")
         self.pipeline = SkyPyPipeline(skypy_config=None, sky_area=sky_area)
 
+    def test_cosmology_initialization(self):
+        # LambdaCDM model
+        cosmo1 = LambdaCDM(H0=70, Om0=0.3, Ode0=0.7)
+        pipeline1 = SkyPyPipeline(cosmo=cosmo1)
+        blue_galaxies = pipeline1.blue_galaxies
+        assert blue_galaxies[0]["z"] > 0
```
sibirrer, line 17: these tests do not test whether the cosmology is overwritten. Are there other test one can do to test a specific cosmology-dependent value?

sibirrer (changes requested):
Thanks a lot!
I suggest you test the cosmology that is being used is the one that you expect it to use, instead of only testing that the code runs.

Then for the formatter, you can locally run
$make format
(see make file for black and docformatter)

trivialTZ:
@sibirrer test for checking 'cosmology is overwritten' is on.

sibirrer:
> @sibirrer test for checking 'cosmology is overwritten' is on.

@trivialTZ I don't see the tests for using the cosmology object of the output of the pipeline result to be compared with the input of the pipeline configuration, or where can I find it?

trivialTZ:
The test between cosmology0 and cosmology1 which has two different
cosmology parameters input and the galaxies number in the output has
significant differences which indicate the change of cosmology.

Simon Birrer ***@***.***>于2023年9月30日 周六18:00写道：

tests/test_Pipelines/test_skypy_pipeline.py
```diff
@@ -8,6 +9,40 @@ def setup_method(self):
...
+        red_galaxies = pipeline0.red_galaxies
+        cosmo0_blue_number = len(blue_galaxies)
+        cosmo0_red_number = len(red_galaxies)
+
+        # LambdaCDM model
+        cosmo1 = LambdaCDM(H0=70, Om0=0.3, Ode0=0.7)
+        pipeline1 = SkyPyPipeline(cosmo=cosmo1)
+        blue_galaxies = pipeline1.blue_galaxies
+        red_galaxies = pipeline1.red_galaxies
+        assert len(blue_galaxies) * 2 < cosmo0_blue_number
```
sibirrer, line 25: this current test does not show that the number of objects with cosmology1 is different than the one with cosmology0

sibirrer (approved):
Looks good. Thank you very much!

PR #76, opened 2023-10-05
Update skypy_pipeline.py

trivialTZ:
access public functions/variables only


sibirrer (approved):
Thank you very much @trivialTZ !

PR #80, opened 2023-10-27
Better Parameter Validation

PatrickRWells:
Hey all,

As promised, here is an implemented system for managing parameter defaults and validating input values. This type of checking is critical for long-running pipelines because it catches mistakes very early in a run and provides a very clear explanation of what is wrong. We want to place clear constraints on what types a parameter can be, what values it can take, and what its defaults are. Importantly, we also want to know exactly where all this information is, so we can update it when needed. This system uses [pydantic](https://github.com/pydantic/pydantic) for parameter validation, which is an industrial-strength standard library for this kind of stuff.

Note: This pull request is also intended as basic documentation for people contributing to slsim, so it'll be fairly long.

The system I've built here is very easy to use, but does require a bit of explanation. Currently, it only works (by design) with inputs to \_\_init\_\_ methods on classes but we can always update that in the future. Importantly, it can be implemented on a **per-class basis.**

To mark the class for validation, we simply decorate it with the @param_check decorator:

`slsim/SomeModule/some_class.py`
```python
from slsim.Utils import param_check

class SomeClass:
    
    @param_check
    def __init__(self, param_a, param_b, param_c):
        self.a = param_a
        self.b = param_b
        self.c = param_c
```

Then create a file called "params.py" in the same folder and add something like the following:

`slsim/SomeModule/params.py`

```python
from pydantic import BaseModel, Field

class SomeClass(BaseModel)
    param_a: list[float]
    param_b: str | list[str] = "some_input"
    param_c: int = Field(default = 2, ge=0)
```
Note that the name of the class in `params.py` must match the name of the original class. We can change this if we want. We just have to pick a convention and stick with it.

This will enforce the following constraints on the parameters:
- param_a must be a list of floats
- parm_b must be a string OR a list of strings. If nothing is provided, the default will be "some_input"
- param_c must be an integer that is greater than 0. If nothing is provided, the default will be 2

**Important note:** All of this is entirely transparent to the user. They simply create an instance of `SomeClass` as they normally would. They will only run into an issue if the parameters they provide are invalid.

All validation is done by `pydantic`, which is incredibly fast. I've updated the GaussianMixtureModel class as an example. The class in the associated `params.py` file looks like this:

`slsim/ParamDistributions/params.py`
```python
from pydantic import BaseModel, PositiveFloat, model_validator
import numpy as np

class GaussianMixtureModel(BaseModel):
    means: list[float] = [0.00330796, -0.07635054, 0.11829008]
    stds: list[PositiveFloat] = [np.sqrt(0.00283885), np.sqrt(0.01066668), np.sqrt(0.0097978)]
    weights: list[PositiveFloat] = [0.62703102, 0.23732313, 0.13564585]
    
    @field_validator("weights")
    @classmethod
    def check_weights(cls, weight_values):
        if sum(weight_values) != 1:
            raise ValueError("The sum of the weights must be 1")
        return weight_values

    @model_validator(mode="after")
    def check_lenghts(self):
        if len(self.means) != len(self.stds) or len(self.means) != len(self.weights):
            raise ValueError("The lenghts of means, stds and weights must be equal")
        return self
```
Note: I just pulled the defaults from the original class definition.

This does three things in addition to simple type checking:
1. Ensures stds and weights are positive values
2. Ensures the total value of the weights == 1
3. Ensures that the length of all three parameters are the same

Some of these were already implemented in the GaussianMixtureModel \_\_init\_\_ function. However, the checking is now much more robust and requires much less code. GaussianMixtureModel now just looks like this
`slsim/ParamDistributions/gaussian_mixture_model.py`
```python

class GaussianMixtureModel:
    """A Gaussian Mixture Model (GMM) class.

    This class is used to represent a mixture of Gaussian distributions, each of which
    is defined by its mean, standard deviation and weight.
    """
    @check_params
    def __init__(self, means, stds, weights):
        """
        The constructor for GaussianMixtureModel class. The default values are the
        means, standard deviations, and weights of the fits to the data in the table
        2 of https://doi.org/10.1093/mnras/stac2235 and others. See "params.py" for
        defaults and validation logic.

        :param means: the mean values of the Gaussian components.
        :type means: list of float
        :param stds: The standard deviations of the Gaussian components.
        :type stds: list of float
        :param weights: The weights of the Gaussian components in the mixture.
        :type weights: list of float
        """
        self.means = means
        self.stds = stds
        self.weights = weights
```

Here's an example of a failure case:

```
>>> from slsim.ParamDistributions.gaussian_mixture_model import GaussianMixtureModel as GMM
>>> GMM(means=[1, 2, 3], stds = [4, 5], weights = [0.4, 0.4, 0.2])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/patrick/code/DESC/slsim/slsim/Util/params.py", line 45, in new_init_fn
    parsed_args = get_defaults(init_fn)(**pargs, **kwargs)
  File "/Users/patrick/opt/anaconda3/envs/dev/lib/python3.10/site-packages/pydantic/main.py", line 164, in __init__
    __pydantic_self__.__pydantic_validator__.validate_python(data, self_instance=__pydantic_self__)
pydantic_core._pydantic_core.ValidationError: 1 validation error for GaussianMixtureModel
  Value error, The lengths of means, stds and weights must be equal [type=value_error, input_value={'means': [1, 2, 3], 'std...ights': [0.4, 0.4, 0.2]}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.4/v/value_error
```

Where we see it is telling us that our three input arrays have to be the same length.

Additional notes:

- We can introduce higher-level validation tools as needed. For example, we could validate that some value has particular units as defined in `astropy.units.` This will require some additional code, but I've basically already written it for `cosmap` so it will be straightforward.
- Related to the above, pydantic is incredibly powerful and has tons of features. We likely won't use most of them, but they're always there. 
- We can eventually extend this system to read information in from configuration files, if we want.
- We can always expand this system to do checkin on more than just initialization functions, but this will definitely introduce some significant complexity on the backend. Happy to do this work though.
- I would strongly recommend we update our minimum Python version to at least 3.10. This allows us to write union types (that is, marking a parameter as allowing multiple types) like `float | int` whereas 3.8 would require `Union[float, int]` and an additional import. Performance has also been a big focus of more recent Python versions, so we're losing out on all that if we stay on 3.8. 
- One thing we may want to change about this code as written is the naming scheme. "Parameter" is the standard name for function inputs, but of course it's also a word we use in a scientific concept. So we may want to pick some convention that avoids confusion.

### DIFF
--- a/slsim/ParamDistributions/gaussian_mixture_model.py
+++ b/slsim/ParamDistributions/gaussian_mixture_model.py
@@ -1,5 +1,5 @@
 import numpy as np
-
+from slsim.Util import check_params
 
 class GaussianMixtureModel:
     """A Gaussian Mixture Model (GMM) class.
@@ -7,12 +7,13 @@ class GaussianMixtureModel:
     This class is used to represent a mixture of Gaussian distributions, each of which
     is defined by its mean, standard deviation and weight.
     """
-
-    def __init__(self, means=None, stds=None, weights=None):
+    @check_params
+    def __init__(self, means, stds, weights):
         """
         The constructor for GaussianMixtureModel class. The default values are the
         means, standard deviations, and weights of the fits to the data in the table
-        2 of https://doi.org/10.1093/mnras/stac2235 and others.
+        2 of https://doi.org/10.1093/mnras/stac2235 and others. See "params.py" for
+        defaults and validation logic.
 
         :param means: the mean values of the Gaussian components.
         :type means: list of float
@@ -21,15 +22,6 @@ class GaussianMixtureModel:
         :param weights: The weights of the Gaussian components in the mixture.
         :type weights: list of float
         """
-        if means is None:
-            means = [0.00330796, -0.07635054, 0.11829008]
-        if stds is None:
-            stds = [np.sqrt(0.00283885), np.sqrt(0.01066668), np.sqrt(0.0097978)]
-        if weights is None:
-            weights = [0.62703102, 0.23732313, 0.13564585]
-        assert (
-            len(means) == len(stds) == len(weights)
-        ), "Lengths of means, standard deviations, and weights must be equal."
         self.means = means
         self.stds = stds
         self.weights = weights
--- a/slsim/ParamDistributions/params.py
+++ b/slsim/ParamDistributions/params.py
@@ -1,0 +1,8 @@
+from pydantic import BaseModel
+import numpy as np
+
+class GaussianMixtureModel(BaseModel):
+    means: list[float] = [0.00330796, -0.07635054, 0.11829008]
+    stds: list[float] = [np.sqrt(0.00283885), np.sqrt(0.01066668), np.sqrt(0.0097978)]
+    weights: list[float] = [0.62703102, 0.23732313, 0.13564585]
+    
--- a/slsim/ParamDistributions/params.py
+++ b/slsim/ParamDistributions/params.py
@@ -1,4 +1,4 @@
-from pydantic import BaseModel, Field, PositiveFloat, model_validator, field_validator
+from pydantic import BaseModel, PositiveFloat, model_validator, field_validator
 import numpy as np
 
 class GaussianMixtureModel(BaseModel):
--- a/slsim/ParamDistributions/params.py
+++ b/slsim/ParamDistributions/params.py
@@ -1,4 +1,5 @@
 from pydantic import BaseModel
+from pydantic import Field
 import numpy as np
 
 class GaussianMixtureModel(BaseModel):
--- a/slsim/ParamDistributions/params.py
+++ b/slsim/ParamDistributions/params.py
@@ -1,12 +1,10 @@
-from pydantic import BaseModel, Field, model_validator
+from pydantic import BaseModel, Field, PositiveFloat, model_validator
 import numpy as np
 
 class GaussianMixtureModel(BaseModel):
     means: list[float] = [0.00330796, -0.07635054, 0.11829008]
-    stds: list[float] = Field(gt=0, 
-                              default=[np.sqrt(0.00283885), np.sqrt(0.01066668), 
-                                       np.sqrt(0.0097978)])
-    weights: list[float] = [0.62703102, 0.23732313, 0.13564585]
+    stds: list[PositiveFloat] = [np.sqrt(0.00283885), np.sqrt(0.01066668), np.sqrt(0.0097978)]
+    weights: list[PositiveFloat] = [0.62703102, 0.23732313, 0.13564585]
     
     @model_validator(mode="after")
     def check_lenghts(self):
--- a/slsim/ParamDistributions/params.py
+++ b/slsim/ParamDistributions/params.py
@@ -14,7 +14,7 @@ class GaussianMixtureModel(BaseModel):
         return weight_values
 
     @model_validator(mode="after")
-    def check_lenghts(self):
+    def check_lengths(self):
         if len(self.means) != len(self.stds) or len(self.means) != len(self.weights):
             raise ValueError("The lenghts of means, stds and weights must be equal")
         return self
--- a/slsim/ParamDistributions/params.py
+++ b/slsim/ParamDistributions/params.py
@@ -1,4 +1,4 @@
-from pydantic import BaseModel, Field, PositiveFloat, model_validator
+from pydantic import BaseModel, Field, PositiveFloat, model_validator, field_validator
 import numpy as np
 
 class GaussianMixtureModel(BaseModel):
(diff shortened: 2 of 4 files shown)


PatrickRWells:
Note: The build failed because it is on Python 3.8, and I used the more robust type hinting conventions that were introduced in Python 3.9. Namely, in Python 3.8 list type hints have to be declared like this:

```python
from typing import List
value: List[float] = [1.0 2.0]
```

While in Python 3.9 the type hinting can be done like this:

```python
value: list[float] = [1.0, 2.0]
```

I think the second is much more natural and requires us to think less, which is an additional reason why I will advocate updating to a newer version of Python. It won't break backward compatibility and just gives us new features.

ajshajib:
@PatrickRWells, thanks a lot for the PR! You can change the Python version [in this file](https://github.com/LSST-strong-lensing/slsim/blob/main/.github/workflows/tests.yml) as part of this PR.

sibirrer:
@PatrickRWells @ajshajib I am a bit worried about the need for requiring 3.9+ Python version. 3.8 should be a long-time supported version and many (or some) machines (be it personal computers or cluster environment) may still run on 3.8. This would narrow down the user base or cause some hassle for some.

PatrickRWells:
@sibirrer @ajshajib  I must admit I really don't see this being an issue when virtual environments exist. I have never met an astronomer who doesn't use Anaconda. Virtual environment support is ubiquitous on clusters. It's not a hassle.

Also, Python 3.8 is [scheduled for end-of-life next year](https://devguide.python.org/versions/). It is not a "long-term supported" version. We should not be building new software with a tool that is about to be abandoned when perfectly good alternatives exist that require virtually no overhead.

sibirrer:
I see @PatrickRWells. I am fine with having a >=3.9 requirement. You can just not assume that all users follow the guidelines and then be too upset if their experience is impacted. At the end it is on the developers to provide a solution to the users. I tend to always favor better backwards compatibility in case the implemented features do not suffer too much and it is not necessary

nkhadka21:
Hi @PatrickRWells , thanks for the PR. You said that this only works for __init__ method. How straightforward is it to extend for other parts of the code? It would be nice if you can write test functions for your implementation.

PatrickRWells:
@sibirrer I think if we're building software with a long-term lifetime in mind, we should try to remain compatible with up-to-date versions of our biggest dependencies. Both numpy and astropy have already dropped support for version 3.8, so we should do the same. Otherwise we run the risk down the line of running into problems (due to bugs or anything else) that are unfixable because we're using an old version, and miss out on updates that may meaningfully improve our experience. I think those teams probably understand the Python ecosystem well enough to make good calls on how much backward compatibility is too much. 

It's also worth mentioning that one of the current big focuses of the Python team is performance, which is a critical aspect of our future work on LSST scale. 3.11 is roughly 30% faster than 3.10, in some applications. It will probably be wise for us to run our code in modern environments anyway.

Fortunately or unfortunately, Python a language under active development, and new versions are expected to be released approximately once a year. I think this is a good thing in the long term, because the language gets meaningfully better. It does require some work on the part of us a developers if we really care about building tools that work for the long term, but that's just part of the deal. 

Changes I will make to the PR:

- I will update the GitHub workflow to do test builds in 3.9-3.11. That will allow us to be confident that our code works well. It is highly unlikely we will run into a scenario where our code does not build in later version.
- I will update requirements.txt to explicitly list 3.9 as the minimum version.

Some things we should consider:

- We should introduce some sort of true dependency management solution. I've used `poetry` quite extensively and found it quite useful (and very easy to use) for pure-python packages. It also can automatically build packages with basically no work required for us as developers. Combined with GitHub actions, we can completely automate releasing builds to PyPI, if we so choose.

PatrickRWells:
> Hi @PatrickRWells , thanks for the PR. You said that this only works for **init** method. How straightforward is it to extend for other parts of the code? It would be nice if you can write test functions for your implementation.

@nkhadka21 Functionally it's not so hard. The back-end machinery for this is quite straightforward. Extending it work with all methods in a class would be straightforward. Extending it to work with functions is a bit more work, but should be doable.

The tricky part here is developer experience. If we're not careful, we run the risk of our "params.py" file getting big and unwieldy, which largely defeats the purpose. One possible approach is to put defaults directly in the same file as the classes/functions.

I think the long-term best solution is to combine this approach with a much tighter API. Right now, this package is mostly a bunch of classes we have to import directly. This is normally how things start off, but we should consider sitting down and really try to properly design an API at some point. This is a big job, but it's the kind of stuff I enjoy working on so I'm happy to do so. But we should do it as a team based on how we expect the package to actually be used.

Changes I will make to this PR:

- I will add test cases for this functionality. I'll make a test class to use, then build out a test suite that checks all the possible different ways people might instantiate an object. Example: all positional arguments, all keyword arguments, a mix, keyword arguments in different orders, etc...

PatrickRWells:
@nkhadka21 @sibirrer 

Updates have been made based on discussion above. Here's a summary

**Small changes:**
- Minimum python version is set to 3.9 for installation
- Testing workflow will now test in python 3.9->3.11 (I have tested on my machine that all tests pass in 3.11 already)
- All code now passes the pre-commit linter workflow

**Bigger Changes:**
- Parameter checking functionality now works with methods besides __init__ and functions. This has required a change in the naming conventions. I have update my original comment on this pull request with the new information.
- This pull request now includes a full test suite for the parameter checking functionality (/tests/test_Params/test_params_check.py)
- I have add validation to the `vel_disp_composite_model` function in `slsim/Deflectors/velocity_dispersion` so I can use it as a test case. Someone who knows more about velocity dispersion stuff may want to look it over and make sure my choices are sanem.

PR #84, opened 2023-11-07
Added movie interpolation + test function

Henry-Best-01:
Added movie interpolation which acts linearly in time. Movie is expected of shape (t, x, y) where the x and y coordinates are generic spatial coordinates. The t value defines the snapshot number. I attempted to keep documentation like the other functions but it still may need tweaking. The time axis must correspond to the Orig_timestamps array, where the simulated times were stored. New_timestamps may be arbitrary. Extrapolation beyond the original bounds is done simply by repeating the boundary values for every pixel.

Also added is a test function which uses the sinusoidal variability as well as generic values to represent "pixel" snapshots at 3 original time steps. The test function interpolates between these values for each pixel at the following important test areas: An endpoint, a midpoint between two points, a central value which is also an original point, and an arbitrary point between an extra long gap (interpolates to 1.5 from original time values 1.0 and np.pi). The test explicitly compares the array created through my function and the expectation calculated pixel by pixel through linear interpolation. I kept it as small as I could.

After downloading the required testing python packages, I received confirmation that it succeeds. I made some plots of it working on significantly larger movies as well (1000 x 400 x 400, summed over pixels in the source plane), which I can share if you want to see it as well.

### DIFF
--- a/slsim/Sources/source_variability/variability.py
+++ b/slsim/Sources/source_variability/variability.py
@@ -44,21 +44,29 @@ def sinusoidal_variability(t, amp, freq):
     """
     return amp * np.sin(2 * np.pi * freq * t)
 
-def interpolate_variability(Movie, Orig_timestamps, New_timestamps):
-    """Interpolates a variable source to any given time series
+def interpolate_variability(Image_series, Orig_timestamps, New_timestamps):
+    """Interpolates a variable source to any given time series.
 
-    :param Movie: 3 dimensional array of shape (t, x, y)
-        relating to snapshots of a variable object in source plane
+    :param Image_series: 3 dimensional array of shape (t, x, y) relating to snapshots of
+        a variable object in source plane
     :param Orig_timestamps: Series of timestamps which represent original snapshots
     :param New_timestamps: Series of new timestamps to interpolate to
-    :return: Linearly interpolated movie along time axis
+    :return: Linearly interpolated image along time axis
     """
-    initial_shape = np.shape(Movie)
+    initial_shape = np.shape(Image_series)
     npix = initial_shape[1] * initial_shape[2]
-    space_positions = np.linspace(1, npix, npix)                        # Order pixels
-    intermediate_movie = np.reshape(Movie, (initial_shape[0], npix))    # Reshape into (t, space) array
-    interpolation = scipy.interpolate.RegularGridInterpolator((Orig_timestamps, space_positions),
-                                        intermediate_movie, bounds_error=False, fill_value=None)
-    new_points = np.meshgrid(New_timestamps, space_positions, indexing='ij')
-    movie_resampled = interpolation((new_points[0], new_points[1]))     # Resample
-    return np.reshape(movie_resampled, (np.size(New_timestamps), initial_shape[1], initial_shape[2]))
+    space_positions = np.linspace(1, npix, npix)  # Order pixels
+    intermediate_movie = np.reshape(
+        Image_series, (initial_shape[0], npix)
+    )  # Reshape into (t, space) array
+    interpolation = scipy.interpolate.RegularGridInterpolator(
+        (Orig_timestamps, space_positions),
+        intermediate_movie,
+        bounds_error=False,
+        fill_value=None,
+    )
+    new_points = np.meshgrid(New_timestamps, space_positions, indexing="ij")
+    Images_resampled = interpolation((new_points[0], new_points[1]))  # Resample
+    return np.reshape(
+        Images_resampled, (np.size(New_timestamps), initial_shape[1], initial_shape[2])
+    )
--- a/slsim/Sources/source_variability/variability.py
+++ b/slsim/Sources/source_variability/variability.py
@@ -1,4 +1,5 @@
 import numpy as np
+import scipy
 
 """This class aims to have realistic variability models for AGN and Supernovae."""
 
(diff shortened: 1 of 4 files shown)


slsim/Sources/source_variability/variability.py
```diff
@@ -1,4 +1,5 @@
 import numpy as np
+import scipy
 
 """This class aims to have realistic variability models for AGN and Supernovae."""
```
ajshajib, line 4:
```suggestion
"""This class aims to have realistic variability models for AGN and supernovae."""
```

ajshajib: Making an insignificant commit to trigger Pre-commit QA actions again.

nkhadka21 (changes requested):

slsim/Sources/source_variability/variability.py
```diff
@@ -42,3 +43,22 @@ def sinusoidal_variability(t, amp, freq):
     :return: variability for the given time
     """
     return amp * np.sin(2 * np.pi * freq * t)
+
+def interpolate_variability(Movie, Orig_timestamps, New_timestamps):
+    """Interpolates a variable source to any given time series
```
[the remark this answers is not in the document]
nkhadka21: Also, check why pre-commit is failing and fix them.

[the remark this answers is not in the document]
Henry-Best-01: It looks like pre-commit was failing because the lines were too long and I had superfluous spacings to align the comments. 

Now, when I run "pre-commit run --show-diff-on-failure --color=always --all-files" locally, I receive all 5 passes, but on Github I'm receiving an error for the hook "black" failing: 
ruff.....................................................................Passed
nbqa-ruff................................................................Passed
black....................................................................Failed
- hook id: black
- files were modified by this hook

reformatted slsim/Sources/source_variability/variability.py
reformatted tests/test_Source/test_source_variability.py

All done! ✨ 🍰 ✨
2 files reformatted, 54 files left unchanged.

black-jupyter............................................................Passed
docformatter.............................................................Passed

 As for why the function needs to be here, I guess it's because it is applicable to any series of screen shots, whether it be AGN or supernovae. Is there a variable class I should put it so it may be accessed cleaner?

ajshajib (changes requested):
@Henry-Best-01 Thanks for the PR! The pre-commit QA is failing for some reason. Can you kindly run these commands in your local `slsim` directory to get rid of the formatting issues and then make a commit?

```
pip install black
pip install docformatter
black .
docformatter ./* -r --black --in-place
```

ajshajib (approved):

slsim/Sources/source_variability/variability.py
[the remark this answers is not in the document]
nkhadka21: Hi @Henry-Best-01 , I am hoping that you have a plan to implement your quasar variability model.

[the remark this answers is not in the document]
Henry-Best-01: Yes I do! It would be silly for me to only contribute only a few lines of code.

I will open a new pull request when I have all the pieces of my black box in place.

tests/test_Source/test_source_variability.py
```diff
@@ -26,6 +27,34 @@ def test_variability_at_t_sinusoidal(self):
         expected_result = np.array([-0.43030122, -0.97536797, -0.14773276])
         npt.assert_almost_equal(result, expected_result, decimal=5)
 
+    def test_interpolation_for_sinusoidal(self):
+        kwargs_model00 = {"amp": 1.0, "freq": 0.5}
```
sibirrer, line 31: can you also describe what you are testing here? This might help the reader and assess the completeness of the test @Henry-Best-01

slsim/Sources/source_variability/variability.py
```diff
@@ -42,3 +43,29 @@ def sinusoidal_variability(t, amp, freq):
     :return: variability for the given time
     """
     return amp * np.sin(2 * np.pi * freq * t)
+
+
+def interpolate_variability(Image_series, Orig_timestamps, New_timestamps):
```
sibirrer, line 48: input variables have conventions of lower_case. Capital letters are reserved for Classes

slsim/Sources/source_variability/variability.py
```diff
@@ -42,3 +43,29 @@ def sinusoidal_variability(t, amp, freq):
...
+    """Interpolates a variable source to any given time series.
+
+    :param Image_series: 3 dimensional array of shape (t, x, y) relating to snapshots of
+        a variable object in source plane
+    :param Orig_timestamps: Series of timestamps which represent original snapshots
```
sibirrer, line 53: It's unclear to me how this input differs from the first axis t in the Image_series. In general, having an explicit definition of these inputs and describing what is happening would be useful. It's unclear to me when reading the docstrings what this function does

sibirrer (changes requested):
Thank you @Henry-Best-01 . I have some requests on the documentation (incl. documentation/description of the tests). Overall, this is a very useful feature. I am just a bit pedantic such that it becomes useful and someone else can also build on it :)
Also, the function/feature you added is not yet accessible through the Variability class. This is ok for now, but when we want to simulate time series, then this needs to be (perhaps for @nkhadka21 )

slsim/Sources/source_variability/variability.py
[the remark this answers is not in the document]
Henry-Best-01: Added a sentence that this will be important for future microlensing implimentation

tests/test_Source/test_source_variability.py
Henry-Best-01: I added details to my tests (all moved into test_param_until)

slsim/Sources/source_variability/variability.py
Henry-Best-01: fixed

slsim/Sources/source_variability/variability.py
Henry-Best-01: Made more clear

slsim/Util/param_util.py
```diff
@@ -85,3 +86,60 @@ def magnitude_to_amplitude(magnitude, mag_zero_point):
...
+    return np.reshape(image_series, (initial_shape[0], number_of_pixels))
+
+
+def pixels_to_images(pixels, original_shape):
+    """Converts a string of pixel snapshots back into a series of image snapshots. This
+    is the inverse of images_to_pixels.
+
+    :param pixels: Series of pixel snapshots to arrange back into the original image
+        shape
+    :param original_shape: The original shape of the image snapshots
```
sibirrer, line 108: I guess this is a tuple? Perhaps specify

sibirrer (approved):
I have one minor comment. I approve it and let @nkhadka21 make the call

slsim/Util/param_util.py
Henry-Best-01: added clarification

nkhadka21 (approved):

PR #87, opened 2023-11-15
Agn sources utility functions

Henry-Best-01:
Added 3 functions to a new file which will contain AGN utility functions. 
These functions: 
 -Determine the innermost stable circular orbit (ISCO) in the Kerr metric around a black hole
 -Calculate the Eddington luminosity for an accretion disk at some given black hole mass (exponent)
 -Calculate the required mass accretion rate for an accretion disk to radiate at a desired Eddington ratio for a given black hole mass (exponent), Eddington ratio, and energy conversion factor.

Also added tests for each function where I compare values of typical and extreme cases, check that errors are returned properly, and astropy units match expectations.


slsim/Util/agn_util.py
```diff
@@ -0,0 +1,61 @@
+import numpy as np
+from astropy import constants as const
+
+
+def spin_to_isco(spin):
+    """Converts dimensionless spin parameter of a black hole to the innermost stable
+    circular orbit (in gravitational radii) [R_g = GM/c^2]
```
nkhadka21, line 7: What is the unit of R_g or dimensionless? It would be good to write the unit.

slsim/Util/agn_util.py
```diff
@@ -0,0 +1,61 @@
...
+    return 3 + z2 - np.sign(spin) * ((3 - z1) * (3 + z1 + 2 * z2)) ** (1 / 2)
+
+
+def calculate_eddington_luminosity(mass_bh_exponent):
+    """Calculates the Eddington luminosity for a black hole mass exponent.
+
+    Eddington_luminosity = 4 * pi * G * mass_bh * mass_proton
+                              * c / sigma_thompson
+
+    :param mass_bh_exponent: solution to log_10(mass_bh / mass_sun).
```
nkhadka21, line 30: Can you make this as a function of black hole mass and it does log10(mass_bh/mass_sun) internally? or maybe you can state this parameter as log of blackhole mass normalized by mass of the sun.

slsim/Util/agn_util.py
```diff
@@ -0,0 +1,61 @@
...
+    return 4 * np.pi * const.G * mass_bh * const.m_p * const.c / const.sigma_T
+
+
+def eddington_ratio_to_accreted_mass(mass_bh_exponent, eddington_ratio, efficiency=0.1):
+    """Calculates the mass that must be accreted by the accretion disk.
+
+    for the accretion disk to radiate at the desired Eddington ratio.
+    Bolometric_luminosity = mass_accreted * c^2 * efficiency
+
+    :param mass_bh_exponent: solution to log_10(mass_bh / mass_sun). Typical AGN have an
```
nkhadka21, line 44: Here also, can you make this as a function of blackhole mass? or maybe just change the description of the mass_bh_exponent.

nkhadka21 (changes requested):
Hi @Henry-Best-01 , thanks alot for your PR! I have few minor comments in your code. Please look at them.

slsim/Util/agn_util.py
Henry-Best-01: Gotcha, R_g is units of length. In general for units I use astropy units (will not come into play until a mass is chosen anyway) so it may be expressed as anything equivalent to meters or centimeters for mks or cgs units.

slsim/Util/agn_util.py
Henry-Best-01: log10(mass_bh/mass_sun) is something I do for ease of use and convenience. Typical values of log10(mass_bh/mass_sun) range from 6.0 to 10.0, are readable and do not run into any issues depending on what units are input. These in mks units are ~ 2 x 10^36 - 2 x 10^40 kg, and in cgs units are ~ 2 x 10^39 - 2 x 10^43 g.  

For units of black hole mass this is 10^6 - 10^8, and I still need to worry about someone tacking on const.M_sun as the units.

I will go with whatever you think is clearest as a user though! As long as I know what checks to put in everything will work out in the end. For now I will adjust the documentation!

Henry-Best-01: I am also changing (bh) to (black_hole) to make it clear

Henry-Best-01:
Sorry for the duplicate files now, I did not want to delete the file which had your comments until you read my responses! the files agn_util.py and test_agn_util.py will be deleted on your confirmation.

nkhadka21:
HI @Henry-Best-01 , thanks for these changes. Please delete these duplicate files.

nkhadka21 (approved):

PR #101, opened 2023-12-01
added noise in static case in a user friendly way!

nkhadka21:


slsim/lsst_science_pipeline.py
```diff
@@ -175,12 +174,13 @@ def lens_inejection_fast(
     lens_pop,
     num_pix,
     delta_pix,
+    mag_zero_point,
```
sibirrer, line 177: add documentation for these newly added parameters

slsim/lsst_science_pipeline.py
```diff
@@ -336,12 +337,13 @@ def multiple_lens_injection_fast(
...
```
sibirrer, line 340: add docstrings

sibirrer (changes requested):
Looks good. Please add some lacking docstrings

Issue #110, opened 2023-12-08
Velocity Dispersion Schechter Integral


P-1884:
There is a difference in the number of galaxies with vel_disp>cutoff when changing the vdmin value in the schechter_vel_disp_redshift and schechter_velocity_dispersion_function integrals. This may be caused by a bug in the normalisation of this integral.

#### Minimal Reproducible Example: 
Changing the vdmin value changes the number of galaxies retrieved with velocity dispersions >200km/s.
<img width="499" alt="image" src="https://github.com/LSST-strong-lensing/slsim/assets/54064261/a5786eaa-77ae-4b9b-96cf-c9377bd08209">

PR #112, opened 2023-12-12
Bug Fix in Schechter Integral

P-1884:
Fix related to issue #110. The integral is made with respect to log(x) rather than x, requiring an additional factor of x=exp(log(x)) when making the change of variables.

This fixes the discrepancy when evaluating for different v_min values - updated reproducible example:
image

### DIFF
--- a/slsim/Deflectors/velocity_dispersion.py
+++ b/slsim/Deflectors/velocity_dispersion.py
@@ -282,7 +282,9 @@ def schechter_vel_disp_redshift(
 
     # gamma function integrand
     def f(lnx, a):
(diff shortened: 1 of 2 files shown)


sibirrer (changes requested):
Thank you very much @P-1884 ! It all looks good. The tests failed as the function changed (to the better). Can you update the test functions such that they pass. You can locally execute the tests with $py.test and can look at the error message on the GitHub actions page. Thank you!

sibirrer (approved):

sibirrer:
Thank you very much @P-1884 ! I'll add you to the list of contributors :)

P-1884:
Thanks so much! :D

PR #142, opened 2024-03-23
Magnitude support for AGN variability

Henry-Best-01:
Working on testing magnitude support for AGN variability.


Henry-Best-01:
I am not sure why codecov/project is failing. If you have any insight, I would appreciate it so I can help fix the problem!

sibirrer:
@Henry-Best-01 hmm, I don't know either. All your changes have been covered with tests, so I think it's good to go from that regard

slsim/Util/astro_util.py
```diff
@@ -555,6 +559,8 @@ def generate_signal(
     high_frequency_slope=3,
     new_mean_amplitude=0,
     new_standard_deviation=1,
+    is_magnitude=False,
```
nkhadka21, line 562: Can you make default is_magnitude = True because magnitude is the conventional unit in SLSim?

slsim/Util/astro_util.py
```diff
@@ -637,6 +675,8 @@ def generate_signal_from_bending_power_law(
     high_frequency_slope=3,
     new_mean_amplitude=0,
     new_standard_deviation=None,
+    is_magnitude=False,
```
nkhadka21, line 678: Here also can you make default is_magnitude = True?

nkhadka21 (changes requested):
Hi @Henry-Best-01 , Thank you very much for this PR! I have few minor comments. Please look at them.

Henry-Best-01:
I changed all references of "is_magnitude" to "normal_magnitude_variance", since the default units are to be magnitude. I also removed "new_" from all user defined means and standard deviations.

nkhadka21 (approved):
Hi @Henery, thank you very much for these changes. It looks good to merge. For another step, I think you can compute light curve in a particular band using mean wavelength of that band (this is a simplest thing we can do).

PR #150, opened 2024-04-03
Composite deflectors

sibirrer:
This PR has an implementation for NFW+Hernquist deflectors and a refactoring of the deflector model structure as a whole.
The PR does not include population level parameters of NFW+Hernquist as this is part of an in-kind contribution.
Tagging here @oguri as well fyi.


slsim/Deflectors/DeflectorTypes/nfw_hernquist.py
```diff
@@ -0,0 +1,84 @@
...
+        :param cosmo: cosmology
+        :type cosmo: ~astropy.cosmology class
+        :return: velocity dispersion [km/s]
+        """
+        # convert radian to arc seconds
+        size_lens_arcsec = self.angular_size_light / constants.arcsec
+
+        m_halo, c_halo = self.halo_properties
+        # convert angular size to physical size
+        lens_cosmo = LensCosmo(z_lens=self.redshift, z_source=10, cosmo=cosmo)
```
nkhadka21, line 35: Do we need to use z_source here? if not, it might be useful to describe why we fix z_source = 10 here.

nkhadka21 (approved):
Hi @sibirrer , thank you very much for this PR! I have a minor comment. Please look at it. Otherwise good to merge.

slsim/Deflectors/DeflectorTypes/nfw_hernquist.py
sibirrer: ah, I only needed this class to get the angular diameter distance to the lens. I simplified it now using directly the astropy cosmology class to get to that

PR #154, opened 2024-04-15
Deleted a coolest test file.

nkhadka21:


sibirrer (commented):
Just to understand, is there another test function that tests this interface in slsim? I think we should have a test function that we can keep track of whether it works or not

sibirrer:
I see @nkhadka21. Can you make an issue such that we can keep track of this? Thank you!

sibirrer (approved):

nkhadka21:
> I see @nkhadka21. Can you make an issue such that we can keep track of this? Thank you!

@sibirrer , coolest addressed my issue and I added coolest_slsim_interface with test functions!

PR #158, opened 2024-04-18
Adding central-object check and DP0 cutout generation

P-1884:
This PR adds two features:

**measure_noise_level_in_RSP_coadd:**
This measures the background noise level from a (presumed) large cutout and returns the 2-sigma upper limit for the total noise in an aperture of specified size. This can be used to exclude coadd cutouts with objects in them. The noise level could vary between coadds so this should be measured on a coadd-by-coadd basis. 

**retrieve_DP0_coadds_from_Rubin_Science_Platform:**
Class to retrieve & save cutouts of coadds, variance maps, PSF arrays and exposure maps from the Rubin Science Platform.

### DIFF
--- a/slsim/lsst_science_pipeline.py
+++ b/slsim/lsst_science_pipeline.py
@@ -836,6 +836,7 @@ def multiple_variable_lens_injection(
         )
     return final_images_catalog
 
+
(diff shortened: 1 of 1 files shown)


nkhadka21 (approved):
Hi @P-1884 , thank you very much for adding these functionality in lsst_science_pipline. These are very useful!

PR #161, opened 2024-04-23
Fix coverage uploading to codecov

ajshajib:


sibirrer (approved):
Thank you @ajshajib !

PR #163, opened 2024-04-25
Added a notebook for image validation.

nkhadka21:
This PR compares image of a DC2 galaxy from slsim and DP0.


slsim/image_simulation.py
```diff
@@ -78,9 +79,28 @@ def sharp_image(
     :param num_pix: number of pixels per axis
     :param with_source: bool, if True computes source
     :param with_deflector: bool, if True includes deflector light
+    :param angle_rot: rotation angle of the coordinate system
```
sibirrer, line 82: provide units [deg]

slsim/image_simulation.py
```diff
@@ -164,34 +185,44 @@ def rgb_image_from_image_list(image_list, stretch):
...
-        pix_center * transform_pix2angle[0, 0] + pix_center * transform_pix2angle[1, 0]
-    )
-    dec_center = (
-        pix_center * transform_pix2angle[0, 1] + pix_center * transform_pix2angle[1, 1]
-    )
-
+    rotation_angle = (
+        angle_rot * np.pi / 180
+    )  # rotation angle in North to West in radian
+    scale1_1 = -np.cos(rotation_angle)
```
sibirrer, line 203: you have quite some duplicating code here to construct the rotation matrix. I'd suggest making an util function for it

slsim/image_simulation.py
```diff
@@ -202,6 +233,7 @@ def image_data_class(
     :param num_pix: number of pixels per axis
     :param transform_pix2angle: transformation matrix (2x2) of pixels into coordinate
         displacements
+    :param angle_rot: rotation angle of the coordinate system
```
sibirrer, line 236: units to be provided

slsim/image_simulation.py
```diff
@@ -78,9 +79,28 @@ def sharp_image(
...
+    scale1_2 = delta_pix * np.sin(rotation_angle)
+    scale2_1 = delta_pix * np.sin(rotation_angle)
+    scale2_2 = delta_pix * np.cos(rotation_angle)
+    ra_at_xy_0_original = (num_pix - 1) / 2
+    dec_at_xy_0_original = (num_pix - 1) / 2
+    pix2angle_transform_rot = np.array([[scale1_1, scale1_2], [scale2_1, scale2_2]])
+    ra_at_xy_0, dec_at_xy_0 = pix2angle_transform_rot.dot(
+        [-ra_at_xy_0_original, -dec_at_xy_0_original]
+    ).T
+    kwargs_pixel_grid = {
```
sibirrer, line 99: Is the issue that the coordinate system is not aligned between DC2 and SLSim or that the angle conventions in the parameters are not aligned? I would not mix the two and make sure that the coordinate system is identical to the definitions in DC2 and instead make sure the parameter definitions are changed. The way you implemented it can lead to confusion when the user also inputs other parts

sibirrer (changes requested):
Thank you @nkhadka21. My main concern is that you are rotating the coordinate system, while the difference might be instead in the convention of the parameters. Please do not mix those two and instead have routines that re-define the parameter conventions instead of compensating with a counter-rotated coordinate system

notebooks/validation_notebooks/image_validation.ipynb
```diff
@@ -0,0 +1,1651 @@
+{
```
sibirrer, line 1: <p>can you come up with some metrics for the validation? Such as a pixel histogram of differences, and that two draws from SLSim noise properties become indistinguishable from the one draw from the DC2 image?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/LSST-strong-lensing/slsim/pull/163/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

nkhadka21:
Hi @sibirrer , I updated this PR addressing all your comments.

sibirrer:
@nkhadka21 so does the coordinate system matches the DC2 conventions in regard to RA and DEC or is the definition of the eccentricities rotated? These are two different things and we should clearly state a) what is the definition of the coordinate system (even without rotation) and then state the parameter definitions in respect to the coordinate system

nkhadka21:
Hi @sibirrer, I changed ellipticity defenition in deflector and source classes. Please look at them.

sibirrer (changes requested):
Hi @nkhadka21, you should not change the functions that output you SLSim conventions, only the ones that explicitly output lenstronomy conventions (as lists of dictionaries), otherwise the conventions might cause confusion

nkhadka21:
Hi @Simon, I changed this! Please look at them.

slsim/Util/param_util.py
```diff
@@ -256,3 +256,15 @@ def deg2_to_cone_angle(solid_angle_deg2):
     solid_angle_sr = solid_angle_deg2 * (np.pi / 180) ** 2
     theta = np.arccos(1 - solid_angle_sr / (2 * np.pi))  # rad
     return theta
+
+
+def ellipticity_slsim_to_lenstronomy(e1_slsim, e2_slsim):
+    """Converts ellipticity component from slsim convension to lenstronomy convention.
```
sibirrer, line 262: do you also want to mention the SLSim conventions and the lenstronomy conventions? Since you know them both it's worth writing them down

sibirrer (approved):
Looks great @nkhadka21. I have one minor suggestion for a better documentation, otherwise good to merge. Thank you very much!

slsim/Util/param_util.py
nkhadka21: I mentioned both slsim and lenstronomy convention for position angle direction.

PR #166, opened 2024-05-15
simpler lens model for gamma=2 setting for speed up in that scenario

sibirrer:
since current power-law slope is by default =2, in that setting using the SIE model is faster to compute

for power-law slope of gamma=2, by default the SIE model is called instead of the EPL model, which can speed up the calculations in that special case.
Unfortunately I had to hack a bit the coolest testing as only the EPL was meant to be tested.
A required lenstronomy update is also needed and the requirements.txt is updated.

### DIFF
--- a/slsim/Deflectors/DeflectorTypes/epl_sersic.py
+++ b/slsim/Deflectors/DeflectorTypes/epl_sersic.py
@@ -73,7 +73,8 @@ class EPLSersic(DeflectorBase):
 
         :return: gamma (with =2 is isothermal)
         """
+
         try:
             return float(self._deflector_dict["gamma_pl"])
-        except:
+        except KeyError:
             return 2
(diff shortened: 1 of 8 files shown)


slsim/Deflectors/DeflectorTypes/epl_sersic.py
```diff
@@ -69,8 +69,11 @@ def light_model_lenstronomy(self, band=None):
...
-        """Properties of the Halo (no halo is defined here)
+        """Mass density logarithmic slope.
 
-        :return: None
+        :return: gamma (with =2 is isothermal)
         """
-        return None
+        try:
+            return float(self._deflector_dict["gamma_pl"])
+        except:
```
nkhadka21, line 78: This is causing pre-commit fail. I think need to specify what are the expected exceptions?

nkhadka21 (changes requested):
@sibirrer , thank you very much for this PR! I noticed a pre-commit fail and I have commented on that particular line.  Please fix it.

nkhadka21 (approved):

PR #167, opened 2024-05-16
bug fix about down-selected source number

sibirrer:
This PR fixes a bug when down-selecting the source galaxy numbers, the strong lensing probability was computed on the total galaxy population while only being drawn from the down-selected one.
In addition I changed some definitions to property returns


sibirrer:
@nkhadka21 there is one line not being tested which stems from the ABC method. If you can review it today that would be great, such that we can merge the fix. Thank you!(Initial report for the issue by @bryce-wedig)

nkhadka21 (approved):

PR #168, opened 2024-05-20
Update source.py

ahuang314:
Modifies source.py to allow kwargs_variability to take roman filters, which eventually get passed into the sncosmo.Models.bandmag function. sncosmo.Models recognizes roman filters

### DIFF
--- a/tests/test_Source/test_source.py
+++ b/tests/test_Source/test_source.py
@@ -181,6 +181,16 @@ class TestSource:
             lightcurve_time=np.linspace(-20, 50, 100),
             cosmo=cosmo,
         )
+        self.source6 = Source(
+            self.source_dict4,
+            variability_model="light_curve",
+            kwargs_variability={"supernovae_lightcurve", "F146"},
+            sn_absolute_mag_band="bessellb",
+            sn_absolute_zpsys="ab",
+            sn_type="Ia",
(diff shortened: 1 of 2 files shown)


sibirrer (changes requested):
Thank you very much @ahuang314! Looks overall great. Can you write a test function importing a Roman band in the Source() class?
Perhaps in the future we can think of the naming of the overall band conventions.

sibirrer (approved):
Thank you @ahuang314 ! I let @nkhadka21 go through it and approve it

tests/test_Source/test_source.py
```diff
@@ -181,6 +181,16 @@ def setup_method(self):
             lightcurve_time=np.linspace(-20, 50, 100),
             cosmo=cosmo,
         )
+        self.source6 = Source(
```
nkhadka21, line 184: Here, you have initiated a source class with roman filter and you need to test some functions. It would be great if you could test this class with test_ps_magnitude_no_variability() and test_ps_magnitude_with_variability() functions. These two functions are the most relavent functions for your changes.

nkhadka21 (changes requested):
@ahuang314 , thank you very much for this PR! I have a comment in testing side. Please address it.

PR #169, opened 2024-05-21
Added Roman Filters to supernovae

ahuang314:
changes to source.py and test_source.py

### DIFF
--- a/tests/test_Source/test_source.py
+++ b/tests/test_Source/test_source.py
@@ -141,6 +141,36 @@ class TestSource:
                 "dec_off",
             ),
         )
+        
+        self.source_dict5 = Table(
+            [
+                [0.5],
+                [1],
+                [4],
+                [0.0002],
+                [0.00002],
+                [0.01],
+                [-0.02],
+                [-0.002],
+                [-0.023],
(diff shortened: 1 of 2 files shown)


nkhadka21 (approved):
@ahuang314 , thank you very much for this PR! Everything looks good to me. I have a minor comment. Please look at it.

tests/test_Source/test_source.py
```diff
@@ -141,6 +141,36 @@ def setup_method(self):
                 "dec_off",
             ),
         )
+
+        self.source_dict5 = Table(
```
nkhadka21, line 145: @ahuang314, you added this source dict but have not used it. I think you can remove this dict.

nkhadka21 (changes requested):

tests/test_Source/test_source.py
ahuang314: Oops, I meant to use that for source6, but since it looks like it is not needed, I removed it

nkhadka21 (approved):
@ahuang314 , thank you very much for quick fix! I everything looks good. congratulation for your first PR merge to slsim.

PR #170, opened 2024-05-21
Added a SupernovaeCatalog class.

nkhadka21:
This PR includes a code to generate a supernovae samples using existing supernovae class and galaxy class. Also, integrates this class with the LensPop class. This class has also renamed Quasars class to PointSources.


slsim/Sources/SupernovaeCatalog/supernovae_sample.py
```diff
@@ -0,0 +1,106 @@
...
+        :return: supernovae host galaxy catalog
+        """
+        pipeline = SkyPyPipeline(
+            skypy_config=self.skypy_config,
+            sky_area=self.sky_area,
+            filters=None,
+            cosmo=self.cosmo,
+        )
+        galaxy_table = pipeline.blue_galaxies
+        galaxy_table_cut = galaxy_table[galaxy_table["z"] <= 0.9329]
```
sibirrer, line 64: why is z<=0.9329 given here? Please document this

slsim/Sources/SupernovaeCatalog/supernovae_sample.py
```diff
@@ -0,0 +1,106 @@
...
+import numpy as np
+
+
+class SupernovaeCatalog(object):
+    """Class to generate a supernovae catalog."""
+
+    def __init__(
+        self,
+        sn_type,
+        band,
```
sibirrer, line 13: is it possible to have the class of the SNe for multiple bands at once? If we do multi-band time series, I see it as a challenge to generate multiple of these classes with exactly the same configurations

slsim/Sources/SupernovaeCatalog/supernovae_sample.py
```diff
@@ -0,0 +1,106 @@
...
+                self.cosmo,
+            )
+            time.append(self.lightcurve_time)
+            magnitude.append(
+                lightcurve_class.get_apparent_magnitude(
+                    self.lightcurve_time, band_string, zpsys=self.mag_zpsys
+                )
+            )
+        if host_galaxy is True:
+            ra_off = np.random.uniform(-5, 5, len(host_galaxies["z"]))
```
sibirrer, line 94: I guess this is pretty arbitrary. Can you document what the intent is of these lines of code and how they need to be replaced in the future with something else?
Alternatively, you can also write a definition with the task of placing the SN in a host galaxy, but then simply write a dummy return to it, but then the design and task of the definition is clear

slsim/lens_pop.py
```diff
@@ -192,6 +192,40 @@ def __init__(
...
+            from slsim.Sources.SupernovaeCatalog.supernovae_sample import (
+                SupernovaeCatalog,
+            )
+
+            for key in kwargs_variability:
+                if key.startswith("ps_mag_"):
+                    suffix = key.split("ps_mag_")[1]
+            supernovae_catalog_class = SupernovaeCatalog(
+                sn_type=sn_type,
+                band=suffix,
```
sibirrer, line 208: here again, might be better to hav a list of bands being passed instead of a single band

slsim/lens_pop.py
```diff
@@ -237,6 +271,36 @@ def __init__(
                     list_type="list",
                     light_profile=source_light_profile,
                 )
+            else:
+                from slsim.Sources.SupernovaeCatalog.supernovae_sample import (
```
sibirrer, line 275: these lines of code following here are very similar to lines of code higher up. Perhaps we can avoid duplications here a bit as this class initialization is rather long

sibirrer (changes requested):
Looks very good. I have some (minor) request for changes but I hope these are able to be addressed. Thank you very much!

nkhadka21:
@sibirrer, I addressed all the comments! Please look at them.

sibirrer (approved):
Looks great, thank you @nkhadka21 !

Issue #156, opened 2024-04-15
method to obtain lenstronomy mass model in Deflector class


furcelay:
I think generating the lenstronomy keywargs for the deflector mass model in the `Lens` class would need one condition for each deflector type and increase complexity over time. A `mass_model_lenstronomy` method (that may take the lens cosmology as an argument) in the `DeflectorBase` class would make that much cleaner.

Issue #134, opened 2024-03-15
Implement Cluster Lens Simulations


furcelay:
The following classes are proposed to implement group/cluster lens simulations within the current SLSIM design. This is only a draft and I really appreciate any comment and contribution.

1. **ClusterPipeline**:
   - **Inherits**: -
   - **Purpose**: Generate a sample of group-to-cluster scale halo parameters (probably NFW) and a sample of  parameters describing the subhalos of each halo (thinking on SIE, ELP or dPIE).
   - **Implementation**: Obtain the mass from a halo mass function, obtain a concentration parameter from a mass-concentration relation, and also an ellipticity. For the subhalo distribution it may use a model (e.g. Oguri's work) or query from a group/cluster catalog (e.g. redMaPPer), assign them to a halo with some criteria (e.g. richness), and fill other information with `SkyPyPipeline`. Maybe use [synth_cluster](https://github.com/LSSTDESC/synth_cluster) for all this (I haven't tested it yet). 

3. **LensClusters** (or maybe a more clear name):
   - **Inherits**: `DeflectorBase`
   - **Purpose**: Handle the parameters of group/cluster dark matter halos and their subhaloes.
   - **Implementation**: This class would use the output of `ClusterPipeline` to build a catalog of individual dark matter halos, encapsulating properties like mass, concentration, number of subhalos, etc. Also may assign the velocity dispersion of the subhalos in the same way as in `EllipticalLensGalaxies` and fill any missing data.  

4. **ClusterLens**:
   - **Inherits**:  `LensedSystemBase`
   - **Purpose**: Combine the parameters of the halo, subhalos, and source into a single lensing system and generate a Lenstronomy model. Consider environmental effects in the same way as in `Lens`. This should also include the surface brightness of every subhalo.
   - **Implementation**: This class would act as a container for a complete lensing system, including a main halo, its subhalos, and the background source. It should handle the combined lensing effect of all these components.

5. **ClusterLensPop**:
   - **Inherits**:  `LensedPopulationBase` 
   - **Purpose**: Contain and manage a population of halos, subhalos, and sources. Be able to generate lensed images that meet some criteria.
   - **Implementation**: This class would generate random combinations of halos, subhalos, and sources to create instances of `ClusterLens`. It should include methods for testing the validity of the generated lens systems.

furcelay:
Anyone interested can follow the progress in this [fork](https://github.com/furcelay/slsim/tree/cluster_lenses)

[...]

PR #171, opened 2024-05-22
Deflector mass method

furcelay:
This pull request adds a new method `mass_model_lenstronomy` to the `DeflectorBase`, `EPLSersic`, `NFWHernquist`, `NFWCluster`, and `Deflector` classes. This method takes a _Lenstronomy_ `LensCosmo` instance and returns the mass profile parameters for the deflector in _Lenstronomy_ conventions.

Although this method depends on a source redshift through the `LensCosmo` instance, it is not tied to a specific source. The results are not stored within the deflector, ensuring that it adheres to the class definition: "Class of a single deflector with quantities only related to the deflector (independent of the source)."

The primary advantage of this addition is the simplification and generalization of the `deflector_mass_model_lenstronomy` method in the `Lens` class. This change removes the need for specific handling of each deflector type, especially the `NFWCluster`, which would otherwise require constructing models for the halo and each subhalo within the function (issue #156). By incorporating this new method, these operations are now managed within the `NFWCluster` and its subhalos, streamlining the process of implementing #134 and other new deflectors.


slsim/Deflectors/DeflectorTypes/deflector_base.py
```diff
@@ -89,6 +89,16 @@ def mass_ellipticity(self):
         )
         return e1_mass, e2_mass
 
+    @abstractmethod
+    def mass_model_lenstronomy(self, lens_cosmo):
+        """Returns lens model instance and parameters in lenstronomy conventions.
+
+        :param lens_cosmo: lens cosmology model
+        :type lens_cosmo: LensCosmo class
```
sibirrer, line 97: can you write this as ~lenstronomy.Cosmo.LensCosmo instance?

sibirrer (approved):
Thank you very much @furcelay! I like this change. It though requires the consistent treatment of the source redshift but I think this is a better way of doing it in the respective classes.
For multiple source redshift, the lenstronomy LensModel() class has a z_source_convention argument and this is the one for which the single-plane equivalent source redshifts are defined in the mass models (so that's what you then need to do, i.e. LensCosmo(z_source=z_source_convention). And then you can have multiple sources with different redshifts appended to it

slsim/Deflectors/DeflectorTypes/deflector_base.py
furcelay: Corrected in the last commit

furcelay:
Thank you @sibirrer! I'll make sure to handle consistent treatment of the source redshift using `z_source_convention` in `LensCosmo` for multiple source cases.

nkhadka21 (approved):
@furcelay , thank you very much for This PR! I like these changes. I approve this PR. Looking forward to seeing your progress in  multiple source with different redshift case.

PR #173, opened 2024-05-24
Integration of NFWCluster into Deflector and Lens classes

furcelay:
This pull request makes three modifications:

* **Integrates `NFWCluster` into `Deflector` class:** 
  1. The `Deflector` class can wrap an `NFWCluster` instance.
  3. Includes a new optional parameter `deflector_kwargs` in `__init__` to pass the `subhalo_list`, a list of deflectors to be treated as subhalos. I decided to make it this way to be general to any deflector that needs something besides `deflector_dict`.
  4. Modifies the `deflector_center` method of `DeflectorBase` to be able to set the position of the subhalos within the cluster in a non-random way by including `center_x` and `center_y` optional keywords in the `deflector_dict`, that are the positions in arcseconds relative to the cluster center. If these keywords are not found it sets a random center.

* **Integrates `NFWCluster` into `Lens` class:** 
   1. The deflector can be initialized as NFWCluster with its subhalos passed as `deflector_kwargs`.
   3. Modified the upper limit when computing the effective Einstein radius on the deflector in `einstein_radius_deflector` method from `rmax=20` to `rmax=50`, this is because cluster lenses tend to have a large Einstein radius. I'm unsure if this impacts the accuracy of the Einstein radius of galaxy lenses, I can evaluate this if needed.
   4. A new test `test_nfw_cluster_lens` verifies that a valid lens can be created with this deflector (in a maximum of 100 trials).
 
* **New notebook**: `cluster_lens.ipynb`  
   1. Uses this code to simulate a cluster lens with an NFW halo and EPL subhalos.
   2. A future scope using the redMaPPer catalog to generate a realistic cluster. Includes new data `redMaPPer/clusters.fits` and `redMaPPer/members.fits` to do so.


sibirrer (approved):
Looks great from my side. I have one comment on the notebook and let @nkhadka21 review it in more detail

furcelay:
<p>I'm doing this right now as the cluster lenses are not yet implemented within the LensPop class, so I need to draw a source through the private attribute _sources externally. This won't be a problem once it is implemented as drawing the source and the deflector would be done inside the LensPop class as is done for other deflectors.</p>

---

 <a href='https://app.reviewnb.com/LSST-strong-lensing/slsim/pull/173/discussion/#comment-2130709143'>View entire conversation</a> on ReviewNB<div id="RAICReplyForNotification-ReviewNBCommentContext-DoNotDelete"></div>

nkhadka21 (changes requested):
@furcelay , thank you very much for this PR! PR looks very good. I have a comment in the notebook, please look at it. Also, you have added two files in data. Can you put only small part of the data here and put full size in data_public repo? especially, members.fits has large size. Also, I noticed a conflict in this PR, please try to solve it from you side.

furcelay:
Hi @nkhadka21! thank you for your revision, I've made the modifications you requested.

nkhadka21 (approved):
@furcelay , thank you very much for quick update! This looks good to me.

PR #174, opened 2024-05-26
convergence of stellar density

sibirrer:
this routine calculates the lensing convergence from the stellar density. It needs to know the total stellar mass of the deflector, the light profile, and then assumes that mass follows light and converts the stellar surface density to mass density, and then lensing convergence.


sibirrer:
@nkhadka21 this is in preparation for the micro-lensing integration such that we can compute the surface mass density of stellar mass

nkhadka21 (approved):
@sibirrer , thank you very much for this PR! It looks good to me.

PR #182, opened 2024-06-12
Co-Moving Density Functions of SN Ia

cmistick:
Added functions to calculate the co-moving number density of SN Ia based on Oguri and Marshall 2010.
- `supernovae_pop.py` in `slsim/slsim/Sources/Supernovae`
      - Includes functions to calculate star formation rate and time delay power distribution.
      - `SNIaRate` class to calculate SN Ia co-moving number density.
- `test_supernovae_pop.py` in `slsim\tests\test_Source`
      - Tests the added functions.


slsim/Sources/Supernovae/supernovae_pop.py
```diff
@@ -0,0 +1,105 @@
...
+"""
+
+
+def calculate_star_formation_rate(z):
+    """Calculates the cosmic star formation rate. (Eq 13 - Oguri and Marshall 2010)
+
+    :param z: redshift
+    :return: cosmic star formation rate in [(h)yr^(-1)Mpc^(-3)]
+    """
+    if z >= 0:
```
sibirrer, line 16: do we need this condition? The reason I am asking is that if not, then this definition can be also executed array-based

slsim/Sources/Supernovae/supernovae_pop.py
```diff
@@ -0,0 +1,105 @@
...
+        star_formation_rate = (0.0118 + 0.08 * z) / (1 + (z / 3.3) ** 5.2)
+        return star_formation_rate
+    else:
+        return np.nan
+
+
+def t_d_power(t_d):
+    """Calculates the power law constraint on time delay. (Eq 14 - Oguri and Marshall 2010)
```
sibirrer, line 24: I would phrase it more general as the delay time distribution, and then stating that it is described as a power-law

slsim/Sources/Supernovae/supernovae_pop.py
```diff
@@ -0,0 +1,105 @@
...
+    def z_from_time(self, t):
+        """Calculates redshift given cosmic time.
+
+        :param t: cosmic time since big bang in [Gy]
+        :param cosmo: cosmology object
+        :type cosmo: ~astropy.cosmology object
+        :return: redshift at time t [float]
+        """
+        if not hasattr(self, "_age_inv"):
+            z_array = np.linspace(0, 1000, 100000)
```
sibirrer, line 55: you might only go to z=100 (fyi, the highest redshift galaxy is at z~14)

sibirrer: do you need 100k interpolation points to be accurate enough? Just asking if we can afford to be a bit less accurate for speed

tests/test_Source/test_supernovae_pop.py
```diff
@@ -0,0 +1,52 @@
...
+    npt.assert_almost_equal(calculate_star_formation_rate(z), 0.0118, decimal=4)
+    z = 3
+    npt.assert_almost_equal(calculate_star_formation_rate(z), 0.1565, decimal=3)
+
+
+def test_t_d_power():
+    t_d = -1
+    npt.assert_equal(t_d_power(t_d), np.nan)
+    t_d = 5
+    npt.assert_almost_equal(t_d_power(t_d), 0.1758, decimal=4)
```
sibirrer, line 27: where does this value come from?

tests/test_Source/test_supernovae_pop.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+def test_z_from_time():
+    z_true = 1
```
sibirrer, line 31: can you test with a few other values in the range [0, 10)?

tests/test_Source/test_supernovae_pop.py
```diff
@@ -0,0 +1,52 @@
...
+    npt.assert_almost_equal(
+        sne_rate._numerator_integrand(t_d, t), 8.9363e-12, decimal=4
+    )
+
+
+def test_calculate_SNIa_rate():
+    z = -1
+    npt.assert_equal(sne_rate.calculate_SNIa_rate(z), np.nan)
+    z = 1.3
+    npt.assert_almost_equal(sne_rate.calculate_SNIa_rate(z), 6.661537e-5, decimal=2)
```
sibirrer, line 52: is this a value that is in quantitative agreement with the number in the literature? Perhaps state from where you have this number you compare to

sibirrer (changes requested):
Looks very good @cmistick ! I have some minor requests and comments

slsim/Sources/Supernovae/supernovae_pop.py
cmistick: If it's fine to have negative redshifts allowed as input (which can result in some formation rates written as complex numbers), then I can remove the condition. Generally, is it fine to assume the user knows this or should I put something in the documentation?

sibirrer: you can state also in the input argument that z>=0 has to be satisfied. Then when the user choses other inputs, the definition can not guarantee any meaningful output

slsim/Sources/Supernovae/supernovae_pop.py
cmistick: I used 100k interpolation points to get more accuracy closer to 0 for the sake of the plots but later added a point to the array well below 0 for this purpose, so I will lower both numbers.

tests/test_Source/test_supernovae_pop.py
```diff
@@ -41,12 +54,13 @@ def test_numerator_integrand():
...
+        sne_rate._numerator_integrand(t_d, t), 0.0002559, decimal=4
     )
 
 
 def test_calculate_SNIa_rate():
     z = -1
     npt.assert_equal(sne_rate.calculate_SNIa_rate(z), np.nan)
-    z = 1.3
-    npt.assert_almost_equal(sne_rate.calculate_SNIa_rate(z), 6.661537e-5, decimal=2)
+    z = 1
```
sibirrer, line 62: can you make tests for a few more redshifts, such as z=0 and perhaps z=2,3?

tests/test_Source/test_supernovae_pop.py
```diff
@@ -1,34 +1,47 @@
...
     calculate_star_formation_rate,
-    t_d_power,
+    delay_time_distribution,
 )
 from slsim.Sources.Supernovae.supernovae_pop import SNIaRate
 from astropy.cosmology import FlatLambdaCDM
 import numpy as np
 import numpy.testing as npt
 
 cosmo = FlatLambdaCDM(70, 0.3)
```
sibirrer, line 10: can you make the tests go in it's now class for tests of class methods? See other tests that do that

sibirrer (changes requested):
looks good. I have some minor requests for testing

tests/test_Source/test_supernovae_pop.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def test_delay_time_distribution():
+    t_d = -1
+    npt.assert_equal(delay_time_distribution(t_d), np.nan)
+    t_d = 5
+    npt.assert_almost_equal(delay_time_distribution(t_d), 5 ** (-1.08), decimal=4)
+
+
+class test_SNIaRate:
```
sibirrer, line 28: classes need to start with capital TestSNIaRate() otherwise they are not being triggered.
Then I suggest you are using a 
def setup_method(self):
definition to set the one_rate and cosmo quantities as self variables within the class

sibirrer (changes requested):
see my comments on the convention of the name of the class method to be recognized such that tests run

cmistick:
At the moment I believe there is an error on one of the lines due to the fact that I had a merge conflict with the auto-fixes from the previous commit I added because I forget to sync them before modifying. I'll re-commit after I have removed it.

tests/test_Source/test_supernovae_pop.py
```diff
@@ -0,0 +1,76 @@
...
+	delay_time_distribution,
+)
+from slsim.Sources.Supernovae.supernovae_pop import SNIaRate
+from astropy.cosmology import FlatLambdaCDM
+import numpy as np
+import numpy.testing as npt
+
+
+def test_calculate_star_formation_rate():
+	z = 0
```
sibirrer: odd formatting

tests/test_Source/test_supernovae_pop.py
```diff
@@ -0,0 +1,76 @@
...
+        self.cosmo = FlatLambdaCDM(H0=70, Om0=0.3)
+        self.z_max = 10
+        self.sne_rate = SNIaRate(
+        	cosmo=self.cosmo,
+        	z_max=self.z_max,
+        )  
+
+    def test_z_from_time(self):
+    	z_true = 0
+		t = self.cosmo.age(z_true)
```
sibirrer: odd formatting, not sure this works

sibirrer (changes requested):
ok, now the formatting is odd. You only need one tab line incline between definition and code

tests/test_Source/test_supernovae_pop.py
cmistick: VSCode was making me revert to an old version, so I copied the code before I did and the spacing got messed up. I fixed it so everything should work now!

cmistick:
There's now an issue with defining the cosmology as I did with skypy.pipeline.

Screenshot 2024-06-12 230327

sibirrer (approved):
Thank you @cmistick ! I approve the PR and let @nkhadka21 go over it as well.
Simon

nkhadka21 (approved):
Hi @cmistick, thank you very much for this PR! The code looks very good to me. I approve this PR.

PR #186, opened 2024-06-18
Added interpolated lightcurves for each image.

nkhadka21:


slsim/lens.py
```diff
@@ -502,6 +531,23 @@ def point_source_magnitude_molet(self, band, time, **kwargs_molet):
 
         # TODO: in what format should be the 2d source profile be stored (as it is time- and wavelength dependent)
         # TODO: do we create full light curves (and save it in cache) or call it each time
+
+        magnitude_without_microlensing = self._point_source_magnitude(
```
sibirrer, line 535: can you make a function that checks whether the interpolation exists and if so, calls the interpolation function to get the correct format back?
And then also define what you need from the molet return to make it an interpolation?
Thank you!

sibirrer (changes requested):

slsim/lens.py
```diff
@@ -502,7 +524,48 @@ def point_source_magnitude_molet(self, band, time, **kwargs_molet):
...
+                ),
+            }
+        ]
+        # using source profile, kappa, gamma, kappa_star, image_observed_time molet
+        # should return lightcurve of each images. The lightcurve can be a interpolated
+        # function or a dictionary of observation time and magnitudes in specified band
+        # as given below.
+        # This molet_output is temporary. Once we call molet, this will be actual molet
+        # output.
+        molet_output = [
```
sibirrer, line 548: can you actually code up to save that as a self variable in cache such that it first used a pre-computed output and only if it does not do that, is going to ask to call molet?

sibirrer (changes requested):
Thank you @nkhadka21. Can you also make a cached version that on the front end, we do not have to worry about running molet multiple times on the same source? Thank you!

slsim/lens.py
nkhadka21: Hi @sibirrer, I made it self variable. Is that all I need to do or I need to check whether these self.molet_output contains the lightcurves or not? If it contains use them otherwise call molet?

slsim/lens.py
```diff
@@ -502,7 +524,50 @@ def point_source_magnitude_molet(self, band, time, **kwargs_molet):
...
+                ),
+            }
+        ]
+        # using source profile, kappa, gamma, kappa_star, image_observed_time molet
+        # should return lightcurve of each images. The lightcurve can be a interpolated
+        # function or a dictionary of observation time and magnitudes in specified band
+        # as given below.
+        # This molet_output is temporary. Once we call molet, this will be actual molet
+        # output.
+        self.molet_output = [
```
sibirrer, line 548: @nkhadka21 I would code it to first check whether this molet output exists, and only if not,, then perform all the parts, and else simply call the interpolation function.
also, please make this variable a private variable

sibirrer (changes requested):

sibirrer (approved):
Thank you @nkhadka21 !

PR #187, opened 2024-06-19
Added lightcone integration of supernovae density

cmistick:
- Added SNeLightcone class in `supernovae_lightcone.py`
  - Utilizes skypy function `redshifts_from_comoving_density` to obtain an array of supernovae redshift locations from SNe comoving density
  - Allows user to pick the time interval and redshift range over which the lightcone is evaluated
  - Can output the redshifts as an astropy table
- Added corresponding TestSNeLightcone class in `test_supernovae_lightcone.py`
- Updated `supernovae_pop.py` to allow array-like inputs to facilitate SNeLightcone class 
- Updated corresponding `test_supernovae_pop.py`

### DIFF
--- a/slsim/Sources/Supernovae/supernovae_lightcone.py
+++ b/slsim/Sources/Supernovae/supernovae_lightcone.py
@@ -27,20 +27,19 @@ class SNeLightcone(object):
         self._time_interval = time_interval
 
         sne_rate = SNIaRate(self._cosmo, self._input_redshifts[-1])
-        h = (self._cosmo.H(0).to_value()/100)
+        h = self._cosmo.H(0).to_value() / 100
 
         self.density = (
             self.convert_density(sne_rate.calculate_SNIa_rate(self._input_redshifts))
-            ) / h
+        ) / h
 
     def convert_density(self, density):
         """Converts SN Ia comoving densities from [yr^(-1)Mpc^(-3)] to have the desired
         time unit.
 
         :param density: initial comoving density of SN Ia [yr^(-1)Mpc^(-3)]
-
         :return: SN Ia comoving density with the desired time unit [day^(-1)Mpc^(-3),
-        hr^(-1)Mpc^(-3), etc.]
+            hr^(-1)Mpc^(-3), etc.]
         """
         time_conversion = (1 * units.year).to(self._time_interval.unit)
         converted_density = density / time_conversion * self._time_interval.value
--- a/slsim/Sources/Supernovae/supernovae_lightcone.py
+++ b/slsim/Sources/Supernovae/supernovae_lightcone.py
@@ -1,0 +1,72 @@
+from skypy.galaxies.redshift import redshifts_from_comoving_density
+from slsim.Sources.Supernovae.supernovae_pop import SNIaRate
+from astropy.table import Table
+from astropy import units
+
+
+class SNeLightcone(object):
+    """Class to integrate SNe comoving density n(z) in light cone volume."""
+
+    def __init__(self, cosmo, redshifts, sky_area, noise, time_interval):
+        """
+        :param cosmo: cosmology object
+        :type cosmo: ~astropy.cosmology object
+        :param redshifts: redshifts for supernovae density lightcone to be evaluated at
+        :type redshifts: array-like
+        :param sky_area: sky area for sampled galaxies in [solid angle]
+        :type sky_area: `~Astropy.units.Quantity`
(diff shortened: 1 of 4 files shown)


slsim/Sources/Supernovae/supernovae_lightcone.py
```diff
@@ -0,0 +1,74 @@
...
+
+        :param density: initial comoving density of SN Ia [yr^(-1)Mpc^(-3)]
+        :return: SN Ia comoving density with the desired time unit [day^(-1)Mpc^(-3),
+            hr^(-1)Mpc^(-3), etc.]
+        """
+        time_conversion = (1 * units.year).to(self._time_interval.unit)
+        converted_density = density / time_conversion * self._time_interval.value
+        return converted_density.value
+
+    def return_supernovae_sample(self):
```
nkhadka21, line 46: Might be good rename it as supernovae_sample().

slsim/Sources/Supernovae/supernovae_lightcone.py
```diff
@@ -0,0 +1,74 @@
...
+        redshift_list = redshifts_from_comoving_density(
+            redshift=self._redshifts,
+            density=self.density,
+            sky_area=self._sky_area,
+            cosmology=self._cosmo,
+            noise=self._noise,
+        )
+        return redshift_list
+
+    def redshift_table(self, redshift_list):
```
nkhadka21, line 62: Instead of making redshift_list an input, you can directly call return_supernovae_sample() function inside this function.

nkhadka21 (changes requested):
Hi @cmistick, thank you very much for this PR! It looks very good. I have minor comments. Please look at them.

slsim/Sources/Supernovae/supernovae_lightcone.py
cmistick: I didn't want to name it the same thing as supernovae_sample.py, but if it won't cause any problems then I'll change it.

slsim/Sources/Supernovae/supernovae_lightcone.py
```diff
@@ -0,0 +1,72 @@
...
+        self._sky_area = sky_area
+        self._noise = noise
+        self._time_interval = time_interval
+
+        sne_rate = SNIaRate(self._cosmo, self._redshifts[-1])
+        self.density = self.convert_density(
+            sne_rate.calculate_SNIa_rate(self._redshifts)
+        )
+
+    def convert_density(self, density):
```
sibirrer, line 34: somewhere the little h got missed, or where do you take it into account?

slsim/Sources/Supernovae/supernovae_lightcone.py
```diff
@@ -0,0 +1,72 @@
...
+        return redshift_list
+
+    def redshift_table(self):
+        """Generates table with redshift locations of supernovae.
+
+        :return: astropy table with redshift locations of supernovae
+        :return type: `~astropy.table.Table`
+        """
+        # Once lightcurve implementation is complete, it will be included here in the table
+        redshift_list = self.supernovae_sample()
```
sibirrer, line 69: could it be that if you execute this routine twice that the same class would return you different supernovae? It might be better to generate the supernovae_sample only once and then re-use it within the same class instance

slsim/Sources/Supernovae/supernovae_pop.py
```diff
@@ -20,14 +21,12 @@ def calculate_star_formation_rate(z):
 def delay_time_distribution(t_d):
     """Calculates the power-law constraint on time delay. (Eq 14 - Oguri and Marshall 2010)
 
-    :param t_d: time delay
+    :param t_d: time delay (t_d>=0)
```
sibirrer, line 24: what is the unit of the delay time?

sibirrer (changes requested):
Looks great @cmistick. I have few minor comments

slsim/Sources/Supernovae/supernovae_lightcone.py
cmistick: For some reason I keep forgetting to put it in the units and then forgot to divide by the factor. I divided by the factor in the initialization function when `self.density` is calculated.

nkhadka21 (approved):
Hi @cmistick, thank you very much for this update! It looks great. I approve and wait for @sibirrer approval.

sibirrer (approved):
Looks great!

PR #188, opened 2024-06-19
Added redshift as an input in the supernovae_catalog() definition of SupernovaeCatalog class.

nkhadka21:


slsim/Sources/SupernovaeCatalog/supernovae_sample.py
```diff
@@ -64,20 +64,36 @@ def host_galaxy_catalog(self):
...
-        """Generates supernovae lightcurves for each host galaxies.
+    def supernovae_catalog(self, redshift=None, host_galaxy=True):
+        """Generates supernovae lightcurves for given redshifts or from host galaxy
+        redshift.
 
         :param host_galaxy: kwargs to decide whether catalog should include host
-            galaxies or not.
-        :return: supernovae catalog
+            galaxies or not. True or False.
+        :return: Astropy Table of supernovae catalog containg redshift, lightcurves,
```
sibirrer, line 73: can you describe a bit more about the light curves?

slsim/lens_pop.py
```diff
@@ -284,8 +284,11 @@ def __init__(
                     sky_area=sky_area,
                 )
                 if source_type == "supernovae":
+                    redshift_sample = supernovae_catalog_class.host_galaxy_catalog()[
```
sibirrer, line 287: will do this do certain calculations twice? Just checking. Perhaps we should also change the source_type for it as this particular type uses a host catalogue, if I am not misunderstood

sibirrer (changes requested):
Thank you @nkhadka21! I have some comments/suggestions for further clarifications

slsim/lens_pop.py
nkhadka21: This does not do any calculation twice. Now, redshift is one of the inputs in supernovae_catalog() function in SupernovaeCatalog class. So, I just get redshift of host galaxies and supplied that as an input but one can supply any list of redshift here. We can replace this line by @cmistick 's redshift sample. This will not use any host galaxy information.

cmistick (approved):

sibirrer (approved):
Looks good

PR #190, opened 2024-06-25
Implemented SN Ia source class in supernovae_sample

cmistick:
- Added SN Ia realistic source class into supernovae_sample so it can be used by LensPop
- Works slowly at the moment due to SNCosmo being called for all SNe rather than only lensed SNe


nkhadka21:
Hi @cmistick, Thank you very much for this PR!  There is a conflict in this PR because of my recent PR merge. Please pull these recent development and probably this will also resolve pytest fail.

nkhadka21 (changes requested):

nkhadka21 (approved):
@cmistick , thank you very much for this PR! it looks very good. I approve this PR.

PR #191, opened 2024-06-25
Few new features in supernovae

nkhadka21:
This PR has implemented following things:
1. Multi-band lightcurves are supported in Source class.
2. connected recent progress in RandomizedSupernova class to the Source, Lens, and LensPop class.
3. Supernovae lightcurves are extracted after getting all the lensed supernovae.


slsim/Halos/halos_lens_base.py
```diff
@@ -496,6 +496,9 @@ def _build_lens_data(self, halos, mass_correction, z1, z2):
         # If this above code need to be changed, notice the change in the following code
         # including the lens_cosmo_dict one since it assume halos is in front of mass sheet
         if not combined_redshift_list.size:
+            warnings.warn(
```
sibirrer, line 499: it looks like you put all these warnings back. Especially this one should not appear

slsim/Sources/SupernovaeCatalog/supernovae_sample.py
```diff
@@ -37,6 +38,8 @@ def __init__(
         :param sky_area: Sky area over which galaxies are sampled. Must be in units of
             solid angle.
         :type sky_area: `~astropy.units.Quantity`
+        :param sn_modeldir: Path to the directory containing supernova files
```
sibirrer, line 41: can you perhaps specify a bit more what type of files these are?

slsim/Sources/SupernovaeCatalog/supernovae_sample.py
```diff
@@ -78,56 +84,63 @@ def supernovae_catalog(self, redshift=None, host_galaxy=True):
...
+                for band in self.band_list:
+                    mag = lightcurve_class.get_apparent_magnitude(
+                        self.lightcurve_time, "lsst" + band, zpsys=self.mag_zpsys
+                    )
+                    getattr(self, f"magnitude_{band}").append(mag)
+            lightcurve_data = {"MJD": time}
+            for band in self.band_list:
+                lightcurve_data["ps_mag_" + band] = getattr(self, f"magnitude_{band}")
+        else:
+            lightcurve_data = {}
```
sibirrer, line 130: do you store at least the inputs that go into RandomizedSupernova() such that when we actually need it down the road, that it can be generated?

sibirrer (changes requested):
Thank you @nkhadka21. I have some minor requests and comments, and looks overall very good

slsim/Halos/halos_lens_base.py
nkhadka21: I synchronized with the TZ's changes.

slsim/Sources/SupernovaeCatalog/supernovae_sample.py
nkhadka21: The only numerical input in the RandomizedSupernova() class is redshift and this is stored. Other inputs are all strings and are available as self variable through LensPop class. These are population level parameters.

sibirrer (approved):

PR #192, opened 2024-06-27
Implementation of large sky area.

nkhadka21:
This PR implements simulation of a lens population in large sky area. We use representative sample of galaxies from a small sky area and scale it to large sky area and simulate a lens population.

### DIFF
--- a/slsim/lens_pop.py
+++ b/slsim/lens_pop.py
@@ -27,9 +27,9 @@ class LensPop(LensedPopulationBase):
         kwargs_mass2light=None,
         skypy_config=None,
         slhammocks_config=None,
+        sky_area=None,
         source_sky_area=None,
         deflector_sky_area=None,
-        full_sky_area=None,
         filters=None,
         cosmo=None,
         source_light_profile="single_sersic",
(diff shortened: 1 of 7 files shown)


slsim/lens_pop.py
```diff
@@ -372,7 +382,7 @@ def deflector_number(self):
 
         :return: number of potential deflectors
         """
-        return self._lens_galaxies.deflector_number()
+        return round(self.factor * self._lens_galaxies.deflector_number())
```
sibirrer, line 385: is there also a deflector_number_selected() argument? Perhaps we missed that here.

sibirrer (changes requested):
as a suggestion, perhaps we can have arguments for 'deflector_sky_area' and 'source_sky_area' as the sky areas to be simulated for the populations, and then the 'lens_sky_area' being the full sky area over which we simulate the population (which can be much larger, but it would not require deflector and sources to have the same sky areas). This might become relevant when i.e. simulating more rare clusters or groups, that require a larger deflector sky area but still a small source sky area.

tests/test_lens_pop.py
```diff
@@ -150,6 +152,7 @@ def test_supernovae_lens_pop_instance():
     assert pes_lens_class._source_type == "point_source"
     assert "z" in pes_lens_class.source.source_dict.colnames
     assert len(pes_lens_class.source.source_dict) == 1
+    assert pes_lens_pop.factor == 1
```
sibirrer, line 155: I don't really see a test of the new feature for the large_skyarea being different than the sky_are, hence I don't see the feature being tested

slsim/lens_pop.py
nkhadka21: In deflector case, self.deflector_number() is the selected number.

nkhadka21:
Hi @sibirrer, I addressed all your comments. its ready for the review.

slsim/lens_pop.py
```diff
@@ -116,9 +132,15 @@ def __init__(
         if deflector_type in ["elliptical", "all-galaxies"] or source_type in [
             "galaxies"
         ]:
-            pipeline = SkyPyPipeline(
+            pipeline_deflector = SkyPyPipeline(
```
sibirrer, line 135: I see that in the case of two different sky areas, SkyPy needs to be run twice. Can you make an exception that when the area of the deflector and sources are the same, that this does not happen?

slsim/lens_pop.py
```diff
@@ -328,8 +352,19 @@ def __init__(
         else:
             raise ValueError("source_type %s is not supported" % source_type)
         self.cosmo = cosmo
-        self.f_sky = sky_area
-
+        self.source_sky_area = source_sky_area
+        self.deflector_sky_area = deflector_sky_area
+        self.f_sky = full_sky_area
+        if full_sky_area is None:
```
sibirrer, line 358: I would change that, such that first the sky areas are set in a certain procedure, and then the factors computed.

slsim/lens_pop.py
```diff
@@ -328,8 +352,19 @@ def __init__(
...
+            self.factor_source = 1
```
sibirrer, line 359: maybe you can keep these factors private variables of the class?

slsim/lensed_population_base.py
```diff
@@ -10,7 +10,9 @@ class LensedPopulationBase(ABC):
 
     def __init__(
         self,
-        sky_area=None,
+        source_sky_area=None,
```
sibirrer, line 13: what happens to this class if the sky areas =None? If the class can not be initialized with the default option, I would ask at least for partial input without default settings. Just checking

sibirrer (changes requested):
Looks great @nkhadka21 . I have some suggestions that perhaps make things backwards compatible and avoid as defaults that the SkyPy catalogue needs to be computed twice.

sibirrer (approved):
Looks great, thank you very much!

PR #193, opened 2024-07-03
Increased Speed of draw_population

cmistick:
- Added speed_parameter to draw_population in LensPop to decrease the number of deflectors drawn and speed up the calculation (test_area is then increased proportionally to account for this change)
- Added test case for the function.


nkhadka21 (approved):
Hi @cmistick , thank you very much for this PR! it looks good to me. I approve it.

PR #194, opened 2024-07-03
updated abundance matching between stellar mass and velocity dispersion for elliptical galaxies

sibirrer:
this PR implements the same abundance matching procedure for elliptical galaxies as it is for all-galaxies, and minor fix in the LOS distribution


nkhadka21 (approved):
Hi @sibirrer , thank you very much for this PR! Looks very good. I approve this PR.

PR #195, opened 2024-07-03
Connected quasar sample to LensPop() class.

nkhadka21:
This PR connects quasar samples drawn from the luminosity function to the LensPop class. Also, I changed some minor things in other part of the code.


slsim/Sources/SupernovaeCatalog/supernovae_sample.py
```diff
@@ -98,7 +98,7 @@ def supernovae_catalog(self, host_galaxy=True, lightcurve=True):
 
             sne_lightcone = SNeLightcone(
                 self.cosmo,
-                redshifts=np.linspace(0, 5.01, 50),
+                redshifts=np.linspace(0, 2.379, 50),
```
sibirrer, line 101: is there a reason to not go to so high redshift? Quasars can be at higher redshift

nkhadka21: Hi @sibirrer , this is for supernovae not for the quasar. @cmistick got this limit but slsim was not updated with this limit.

sibirrer (approved):

PR #209, opened 2024-07-18
Implemented ellipticity dependence in SN Ia host galaxy offsets

cmistick:
- Added ellipticity dependence in the offset calculation within  `supernovae_host_galaxy_offset()`
    - Utilized `galaxy_projected_eccentricity()` from `galaxies.py` and `transform_e1e2_product_average()` from Lenstronomy's `param_util.py`
- Added more comments in `supernovae_host_galaxy_offset()`


cmistick:
Here is how the distribution of offsets appears when this PR's changes are used (1 deg^2 & 2597 SN-host pairs):
offset_dist_elliptical

sibirrer:
@nkhadka21: Carly implemented a elliptical distortion of the positions to better monique the elliptical light profile distributions. Chances are that the e1, e2 definitions are not identical with lenstronomy's one, and I advised Carly to use the lenstronomy function to displace/distort the SNe. Is there a check we can do to make sure we are doing things consistently?

slsim/Sources/SupernovaeCatalog/supernovae_sample.py
```diff
@@ -30,27 +33,55 @@ def supernovae_host_galaxy_offset(host_galaxy_catalog):
 
     offsets = []
     position_angle = []
+    e1 = []
```
nkhadka21, line 36: I think e1 and e2 should be return from this function and store them in host galaxy catalog. Otherwise, ellipticity component for each galaxy will be calculated in Galaxy class and it will not be consistent with e1 and e2 used here.

slsim/Sources/SupernovaeCatalog/supernovae_sample.py
```diff
@@ -30,27 +33,55 @@ def supernovae_host_galaxy_offset(host_galaxy_catalog):
...
-    dec_off = dec_off.to(units.arcsec)
+    transformed_ra_off = []
+    transformed_dec_off = []
+
+    # Transform the offset coordinates with eccentricities e1, e2 into elliptical coordinate system
+    for i in range(len(host_galaxy_catalog)):
+        ra_off, dec_off = transform_e1e2_product_average(
+            original_ra_off[i],
+            original_dec_off[i],
+            e1[i],
```
nkhadka21, line 76: For slsim to lenstronomy conversion of ellipticities, use this function: https://github.com/LSST-strong-lensing/slsim/blob/c6b3fa82eae8a0e58e7ec539d69ce328a7a30689/slsim/Util/param_util.py#L261 .

nkhadka21 (changes requested):
Hi @cmistick , thank you very much for this PR! This looks wonderful. I have minor comments. Please look into them.

slsim/Sources/SupernovaeCatalog/supernovae_sample.py
cmistick: Do you mean that the function should return both ra and dec offsets, as well as e1 and e2? Or should it be split into 2 different functions, one for calculating ra and dec offsets and one for calculating e1 and e2?

nkhadka21: I think function should return both ra, dec and e1, e2 so that you can use e1 and e2 in host galaxy catalog.

cmistick: Sounds great. That's what I chose to do in the latest commit!

slsim/Sources/SupernovaeCatalog/supernovae_sample.py
```diff
@@ -30,27 +36,61 @@ def supernovae_host_galaxy_offset(host_galaxy_catalog):
...
-    dec_off = dec_off.to(units.arcsec)
+    transformed_ra_off = []
+    transformed_dec_off = []
+    lenstronomy_e1 = []
+    lenstronomy_e2 = []
+
+    # Transform the offset coordinates with eccentricities e1, e2 into elliptical coordinate system
+    for i in range(len(host_galaxy_catalog)):
+        temp_e1, temp_e2 = ellipticity_slsim_to_lenstronomy(e1[i], e2[i])
+        lenstronomy_e1.append(temp_e1)
```
nkhadka21, line 79: do you need to append them? I think you can directly use temp_e1 and temp_e2 (perhaps remane to lenstronomy_e1 and lenstronomy_e2) in transform_e1e2_product_average() function.

nkhadka21 (commented):
Hi @cmistick , thank you very much for this update! looks very good. I have a minor comment. Please look into the comment.

nkhadka21 (approved):
Hi @cmistick , thank you very much! looks very good.

PR #216, opened 2024-07-31
Adding an initialization class for agn objects to be used with the so…

Henry-Best-01:
This is the AGN object which will be called with random parameters in order to populate a source catalog. For now, it only has a method to calculate the (not K-corrected) mean magnitudes from other bands. I may have to add another couple methods afterwards in order to connect this AGN object to the source module.


sibirrer:
@Henry-Best-01 is this ready for review? Just checking

Henry-Best-01:
Not yet, sorry! 

I am in the middle of connecting the agn object to both the variability class and the source class. Then I will make sure it is working as expected!

I think I am almost done with the code, but testing will be important. The time frame is uncertain based on how many bugs I find with it... but I plan to start testing today.

Henry-Best-01:
Hey @nkhadka21 , I've been trying to fix some bugs with making these connections, but I keep getting failures on some supernova sources. I was trying to add dummy values in the source_dict for each test case as a hotfix, but it is still failing. 

I did my best to not touch the code involved with other sources and only include what I needed to get a variable AGN source included, but I can't seem to get past these errors. 

In the mean time, I will write some unrelated tests for the AGN cases.

slsim/Sources/source.py
```diff
@@ -68,20 +74,25 @@ def kwargs_variability_extracted(self):
             ##Here we prepare variability class on the basis of given
             # kwargs_variability.
             kwargs_variab_extracted = {}
-            kwargs_variability_list = ["supernovae_lightcurve"]
```
nkhadka21, line 71: I suggest to use another kwargs_variability_list for agn. Let's say kwargs_variability_list_agn = ["agn_lightcurve"]. Then, write elif statement for agn with out touching if statement. This will give us waht is happening here.

nkhadka21 (commented):
Hi @Henry-Best-01 , thank you very much for this PR! I have one suggestion that might be helpful to avoid test error.

slsim/Sources/source.py
Henry-Best-01: Got it sorted! Apparently I added an indent to one of the loops.

Henry-Best-01:
I think there's still a bug which I will fix tomorrow, which is causing one of the loops to not enter. This is the line 164 of source.py which is not covered by the test, even though my test should be covering this.

nkhadka21:
> I think there's still a bug which I will fix tomorrow, which is causing one of the loops to not enter. This is the line 164 of source.py which is not covered by the test, even though my test should be covering this.

Hi @Henry-Best-01 , thank you very much for the update! It seems like line 164 is a blank space! There is no code.

Henry-Best-01:
Oh haha, I was trying to get it to get into that loop, but I guess it was!

slsim/Sources/source.py
```diff
@@ -125,6 +137,80 @@ def kwargs_variability_extracted(self):
...
+                    # Populate "None" for optional keys related to drawing random AGN
+                    random_kwargs = {"random_seed": None, "input_agn_bounds_dict": None}
+                    for opt_key in random_kwargs.keys():
+                        if opt_key in agn_kwarg_dict.keys():
+                            random_kwargs[opt_key] = agn_kwarg_dict[opt_key]
+                            del agn_kwarg_dict[opt_key]
+
+                    # Create the agn object
+                    self.agn_class = agn.RandomAgn(
+                        self.source_dict["i_band_mag"],
```
nkhadka21, line 167: quasar source dict have "ps_mag_i" for magnitude name convention. Here also do "ps_mag_" + band, and band can be passed through kwargs_variability as in supernovae case. kwargs_varibility is already connected through Lens and LensPop class so user can provide any keyword argument for variability model using this dictionary.

slsim/Sources/source.py
```diff
@@ -125,6 +137,80 @@ def kwargs_variability_extracted(self):
...
+                        input_agn_bounds_dict=random_kwargs["input_agn_bounds_dict"],
+                        **agn_kwarg_dict
+                    )
+
+                    # Get mean mags across LSST filters
+                    # Fix to eventually give the user power to choose other surveys
+                    mean_magnitudes = self.agn_class.get_mean_mags("lsst")
+
+                    # These are lsst filters, hard coded in for now. Fix later.
+                    filters = ["u", "g", "r", "i", "z", "y"]
```
nkhadka21, line 180: you can demand to pass these filters through kwargs_variability and can be used whenever needed. We have done this in supernovae case.

slsim/Sources/source.py
```diff
@@ -125,6 +137,80 @@ def kwargs_variability_extracted(self):
...
+                        self.agn_class.variable_disk.driving_signal_kwargs[
+                            "mean_magnitude"
+                        ] = mean_magnitudes[index]
+
+                        reprocessed_lightcurve = reprocess_with_lamppost_model(
+                            self.agn_class.variable_disk
+                        )
+
+                        times = reprocessed_lightcurve["MJD"]
+                        magnitudes = reprocessed_lightcurve["ps_mag_lsst2023-" + band]
```
nkhadka21, line 200: I suggest to use reprocessed_lightcurve["ps_mag_" + band]

slsim/Sources/source.py
```diff
@@ -125,6 +137,80 @@ def kwargs_variability_extracted(self):
...
+
+                        new_column = Column([float(min(magnitudes))], name=filter_name)
```
nkhadka21, line 202: In case of supernovae, supernovae sample does not have ps_mag_i or simply peak magnitude so we store them from the lightcurve. But, in quasar case, quasar sample have ps_mag_i and you use this magnitude from the source _dict while getting the quasar lightcurve. So, I think it is not necessary to store peak magnitude from the lightcurve and lines 202 to 206 can be excluded.

slsim/Sources/source.py
```diff
@@ -125,6 +137,80 @@ def kwargs_variability_extracted(self):
...
+                z = self.source_dict["z"]
+                if self.cosmo is None:
+                    raise ValueError(
+                        "Cosmology cannot be None for AGN class. Please"
+                        "provide a suitable astropy cosmology."
+                    )
+
+                else:
+                    # Get the dict object from the astropy.table.column
+                    agn_kwarg_dict = self.source_dict["kwargs_agn_model"].data[0][0]
```
nkhadka21, line 156: You can demand kwargs_agn_model keywords from the kwargs_variability (this is where user should put all the keywords for variability model) and based on those keywords construct agn_kwargs_dict using source_dict where the values of these keyswords should be available. In source_dict, these values should be placed in different column instead of a compact dictionary.

nkhadka21 (changes requested):
Hi @Henry-Best-01 , thank you very much for this wonderful work! This looks good to me. But, I have some comments on design aspect. Please look at these comments and please feel free to reach out to me if my comments are not clear.

slsim/Sources/agn.py
```diff
@@ -0,0 +1,221 @@
...
+        'black_hole_spin': spin of the SMBH 'inclination_angle': inclination of the AGN
+        disk in degrees 'r_out': Maximum radius of the disk in gravitational radii
+        'r_resoultion': Number of pixels the disk is resolved to 'eddington_ratio':
+        fraction of the eddington luminosity the disk is radiating with
+        'accretion_disk': accretion disk model
+    """
+
+    def __init__(
+        self,
+        i_band_mag,
```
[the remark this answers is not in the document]
Henry-Best-01: Updated the function to incorporate any band, along with an additional test

slsim/Sources/source.py
Henry-Best-01: Got it! I just copied these lines of code thinking you had to create a new Column for each band

slsim/Sources/agn.py
```diff
@@ -0,0 +1,221 @@
...
+            self.kwargs_model["time_array"] = self.kwargs_model[
+                "intrinsic_light_curve"
+            ]["MJD"]
+            self.kwargs_model["magnitude_array"] = self.kwargs_model[
+                "intrinsic_light_curve"
+            ]["ps_mag_intrinsic"]
+
+        self.variable_disk = Variability("lamppost_reprocessed", **self.kwargs_model)
+
+    def get_mean_mags(self, survey):
```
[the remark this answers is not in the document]
Henry-Best-01: Got it!

slsim/Sources/source.py
Henry-Best-01: Got it! 

I wrote a method at the bottom of source.py to contain the extraction of AGN parameters.

slsim/Sources/source.py
Henry-Best-01: I did have to include the creation of the new columns so that the method point_source_magnitude could find the column named "ps_mag_"+band.

slsim/Sources/source.py
```diff
@@ -140,6 +169,121 @@ def kwargs_variability_extracted(self):
...
+                        )
+
+                        # Prepare the light curve to be extracted
+                        times = reprocessed_lightcurve["MJD"]
+                        magnitudes = reprocessed_lightcurve[
+                            "ps_mag_" + speclite_names[index]
+                        ]
+
+                        # Prepare the time variable magnitude
+                        new_column = Column([min(magnitudes)], name=filter_name)
```
nkhadka21, line 270: agn class uses ps_mag_i from the source_dict. Then, it gives a lightcurve in i-band. Does peak magnitude of this lightcurve match the ps_mag_i that you used in agn class? ps_mag_i that is provided in source_dict is a peak magnitude of the quasar. If ps_mag_i and min(magnitudes) are suppose to be the same quantity, I think we should not replace ps_mag_i by min(magnitudes).

nkhadka21 (changes requested):
Hi @Henry-Best-01 , thank you very much for this update! We are very close to the completion. I have some minor comments/questions. Please look at them and if possible address them.

nkhadka21 (approved):
Hi @Henry-Best-01 , thank you vefry much for this wonderful work! It looks great. I approve it.

PR #217, opened 2024-07-31
Point source cut

nkhadka21:
This PR has a feature to apply cuts in point source magnitude and redshift.


slsim/Sources/point_sources.py
```diff
@@ -24,6 +27,8 @@ def __init__(
         :param sky_area: Sky area over which galaxies are sampled. Must be in units of
             solid angle.
         :type sky_area: `~astropy.units.Quantity`
+        :param kwargs_cut: cuts in parameters: band, band_mag, z_min, z_max
```
sibirrer, line 30: here you can specify that these are arguments that go into the deflector_cut() definition

sibirrer (changes requested):
Looks good. I have only one minor ask for the documentation. Haven't checked whether tests are passing

sibirrer (approved):

PR #218, opened 2024-08-01
Added a feature in lsst_science_pipeline module to save data while runing lens injection

nkhadka21:
This PR has a feature which allow user to save injected image catalog while injection process is in progress. One need to pass path where they want to save the data.


slsim/lsst_science_pipeline.py
```diff
@@ -749,6 +762,40 @@ def dp0_time_series_images_data(butler, center_coord, radius="0.1", band="i", si
     return table_data
 
 
+def multiple_dp0_time_series_images_data(
+    butler, center_coords_list, radius="0.034", band="i", size=101, output_file=None
+):
+    """Creates multiple time series data from dp0 data.
```
sibirrer, line 768: can you elaborate a bit what 'multiple' means. Multiple bands? Multiple objects with different cutouts?

slsim/lsst_science_pipeline.py
```diff
@@ -772,7 +819,17 @@ def variable_lens_injection(
         in days for each single exposure images in time series images)
     :return: Astropy table of injected lenses and exposure information of dp0 data
     """
-
+    ##the range of observation time of single exposure images might be outside of the
```
sibirrer, line 822: double ##

tests/test_lsst_science_pipeline.py
```diff
@@ -29,6 +29,7 @@ def pes_lens_instance():
             variability_model="sinusoidal",
             kwargs_variability={"amp", "freq"},
             cosmo=cosmo,
+            lightcurve_time=np.linspace(0, np.pi, 100),
```
sibirrer, line 32: I don't understand why lightcurve_time is linear between zero and pi

sibirrer (changes requested):
Looks very good. I have some minor request

tests/test_lsst_science_pipeline.py
nkhadka21: Here, I am using sinusoidal function as a lightcurve for testing purpose. So, I just provided values between 0 to pi as a time array. This is a old test function (we are not writing any test function for lsst_science_pipeline. These are the only function we can test). If we use actual supernovae or quasar lightcurves, then we should provide days as observation time.

sibirrer (approved):

PR #222, opened 2024-08-03
Added factor for comoving SN Ia density

cmistick:
- Added 1/(1+z)^3 factor in SNeLightcone class density conversion calculation
- Updated test_SNeLightcone class

### DIFF
--- a/slsim/Sources/Supernovae/supernovae_lightcone.py
+++ b/slsim/Sources/Supernovae/supernovae_lightcone.py
@@ -28,6 +28,7 @@ class SNeLightcone(object):
         sne_rate = SNIaRate(self._cosmo, self._input_redshifts[-1])
         h = self._cosmo.H(0).to_value() / 100
 
+        # Account for included factor of h and convert the density from physical to comoving
         self.density = (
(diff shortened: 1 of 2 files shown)


slsim/Sources/Supernovae/supernovae_lightcone.py
```diff
@@ -30,7 +30,7 @@ def __init__(self, cosmo, redshifts, sky_area, noise, time_interval):
 
         self.density = (
             self.convert_density(sne_rate.calculate_SNIa_rate(self._input_redshifts))
-        ) / h
+        ) / (h * (1 + self._input_redshifts) ** 3)
```
nkhadka21, line 33: Might be good to write why you devided by (1+self._input_redshifts)**3 factor.

nkhadka21 (changes requested):
Hi @cmistick , thank you for this quick fix! I have a very minor comment. Please address that and I will merge it.

nkhadka21 (approved):
Hi @cmistick , thank you very much for quick update! Looks great. I approve it.

PR #231, opened 2024-08-06
Added point source magnitude cut consistently through all cases.

nkhadka21:
This PR has implemented point source magnitude cuts consistently in all cases. Also, made some minor changes in quasar_pop.


sibirrer (approved):
Looks great. Thank you @nkhadka21 !

PR #232, opened 2024-08-07
Added SN Ia luminosity function calculation

cmistick:
- Added luminosity function calculation (OM10 - Eq. 18) within `supernovae_pop.py`
- Added corresponding test function in `test_supernovae_pop.py`

### DIFF
--- a/slsim/Sources/Supernovae/supernovae_pop.py
+++ b/slsim/Sources/Supernovae/supernovae_pop.py
@@ -54,6 +54,7 @@ class SNIaRate(object):
         """Calculates redshift given cosmic time.
 
         :param t: cosmic time since big bang in [Gy]
+
         :return: redshift at time t [float]
         """
         if not hasattr(self, "_age_inv"):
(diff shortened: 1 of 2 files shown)


slsim/Sources/Supernovae/supernovae_pop.py
```diff
@@ -99,3 +100,23 @@ def calculate_SNIa_rate(self, z, eta=0.04):
             SNIa_rate_list.append(SNIa_rate)
 
         return np.array(SNIa_rate_list)
+
+    def calculate_luminosity(self, M, z):
```
nkhadka21, line 104: change this name to luminosity_function().

nkhadka21 (changes requested):
Hi @cmistick , thank you very much for this PR! the function looks good. You can use this function to sample supernovae magnitudes. For this, you need to compute cdf using luminosity function and sample magnitudes from the cdf. Then, you should get supernovae redshift and magnitudes in your sample. Please look at @emsonmez 's class for the detailed process.

PR #234, opened 2024-08-09
Fixed some typo and changed some code for consistency.

nkhadka21:
This PR fixes a typo in the SNIa number density. Additionally, in the quasar case, the physical-to-comoving volume factor was placed in the luminosity function and removed from the density. The density is derived from the luminosity function, so this factor consistently applies to the density.

### DIFF
--- a/slsim/Sources/QuasarCatalog/quasar_pop.py
+++ b/slsim/Sources/QuasarCatalog/quasar_pop.py
@@ -131,7 +131,9 @@ class QuasarRate(object):
 
     def dPhi_dM(self, M, z_value):
         """Calculates dPhi_dM for a given M and redshift according to Eq (10) in Oguri &
-        Marshall (2010): DOI: 10.1111/j.1365-2966.2010.16639.x.
+        Marshall (2010): DOI: 10.1111/j.1365-2966.2010.16639.x. The (1 + z)^-3 factor 
+        converts from physical to comoving volume to account for the expansion of the 
+        universe.
 
         :param M: Absolute i-band magnitude.
         :type M: float or numpy.ndarray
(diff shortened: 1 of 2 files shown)


slsim/Sources/QuasarCatalog/quasar_pop.py
```diff
@@ -131,7 +131,9 @@ def M_star(self, z_value):
 
     def dPhi_dM(self, M, z_value):
         """Calculates dPhi_dM for a given M and redshift according to Eq (10) in Oguri &
-        Marshall (2010): DOI: 10.1111/j.1365-2966.2010.16639.x.
+        Marshall (2010): DOI: 10.1111/j.1365-2966.2010.16639.x. The (1 + z)^-3 factor
+        converts from physical to comoving volume to account for the expansion of the
+        universe.
 
         :param M: Absolute i-band magnitude.
         :type M: float or numpy.ndarray
```
sibirrer, line 139: can you explicitly specify whether the return units are physical or co-moving?

sibirrer (changes requested):

cmistick (approved):
Looks good! I forgot about the h factor being incorrect, thanks for the fix!

slsim/Sources/QuasarCatalog/quasar_pop.py
emsonmez, line 1: LGTM! I guess this is needed to make sure that our dPhi_dM graph is consistent with Figure 1 of OM10.

emsonmez (approved):
LGTM! Although I would still believe that applying the factor during the integration in `n_comoving` is better than before integration in `dPhi_dM`.

sibirrer (approved):

Issue #224, opened 2024-08-06
Refactoring SLSim - Simplify LensPop


jacob-hjortlund:
## Current Status

Currently `LensPop` takes strings and dictionaries as inputs at initialization, and uses `if`-statements "under the hood" to select what objects are initialized for the deflector and source populations. This hard-coding reduces legibility and extensibility, leading contributors to have to modify the `LensPop` if they modify or add population objects.

## Proposed Solution

Follow the dependency injection design pattern and have `LensPop` take a lens and source population as input, with `LensPop` expecting certain attributes and methods be available from the provided class instances. This reduces the scope / responsibilities of  `LensPop` while allowing future contributors to create drop-in substitutions, as long as they fullfill the interface expected by `LensPop`, see Figure in #223. 

## To-Do List

- [ ] Move `SkyPyPipeline` out of `LensPop`.
- [ ] Move instances of `_lens_galaxies` attribute (i.e instances of `DeflectorBase` subclasses) out of `LensPop` and instead expect as input to `LensPop`.
- [ ] Move instances of `_sources` attribute (i.e instances of `SourcePopBase`) out of `LensPop` and instead expect as input to `LensPop`.
- [ ] Update unit tests.

Issue #225, opened 2024-08-06
Refactoring SLSim - Merging DeflectorBase and SourcePopBase into single Population class.


jacob-hjortlund:
## Current Status

`DeflectorBase` and `SourcePopBase` currently share the same overall responsibility, providing a sample deflector or source from a predefined population of galaxies.

## Proposed Solution

Merge `DeflectorBase` and `SourcePopBase` into a single `Population` class that takes an `AstroPy.Table` as input and provides a method for drawing a single sample from the given table, see Figure in #223. In the future this class should also accept transient metadata associated with the galaxies and provide them with sampling, as well as construct a `Galaxy` object from the drawn sample. 

## To-Do List

- [ ] TBD

PR #236, opened 2024-08-14
 Refactoring SLSim - Simplify LensPop

jacob-hjortlund:
WIP pull request for #224 . 

## To-Do List

- [x] Move `SkyPyPipeline` out of `LensPop`.
- [x] Move instances of `_lens_galaxies` attribute (i.e instances of `DeflectorBase` subclasses) out of `LensPop` and instead expect as input to `LensPop`.
- [x] Move instances of `_sources` attribute (i.e instances of `SourcePopBase`) out of `LensPop` and instead expect as input to `LensPop`.
- [x] Update unit tests.
- [ ] Update tutorials to reflect changes.

### DIFF
--- a/slsim/Deflectors/__init__.py
+++ b/slsim/Deflectors/__init__.py
@@ -1,5 +1,7 @@
-from .all_lens_galaxies import AllLensGalaxies
-from .compound_lens_halos_galaxies import CompoundLensHalosGalaxies
-from .elliptical_lens_galaxies import EllipticalLensGalaxies
-from .deflector import Deflector
-from .deflectors_base import DeflectorsBase
+from .all_lens_galaxies import AllLensGalaxies as AllLensGalaxies
+from .compound_lens_halos_galaxies import (
+    CompoundLensHalosGalaxies as CompoundLensHalosGalaxies,
+)
+from .elliptical_lens_galaxies import EllipticalLensGalaxies as EllipticalLensGalaxies
+from .deflector import Deflector as Deflector
+from .deflectors_base import DeflectorsBase as DeflectorsBase
--- a/slsim/Deflectors/__init__.py
+++ b/slsim/Deflectors/__init__.py
@@ -1,7 +1,13 @@
-from .all_lens_galaxies import AllLensGalaxies as AllLensGalaxies
-from .compound_lens_halos_galaxies import (
-    CompoundLensHalosGalaxies as CompoundLensHalosGalaxies,
-)
-from .elliptical_lens_galaxies import EllipticalLensGalaxies as EllipticalLensGalaxies
-from .deflector import Deflector as Deflector
-from .deflectors_base import DeflectorsBase as DeflectorsBase
+from .all_lens_galaxies import *
+from .compound_lens_halos_galaxies import *
+from .elliptical_lens_galaxies import *
+from .deflector import *
+from .deflectors_base import *
+
+# from .all_lens_galaxies import AllLensGalaxies as AllLensGalaxies
+# from .compound_lens_halos_galaxies import (
+#     CompoundLensHalosGalaxies as CompoundLensHalosGalaxies,
+# )
+# from .elliptical_lens_galaxies import EllipticalLensGalaxies as EllipticalLensGalaxies
+# from .deflector import Deflector as Deflector
+# from .deflectors_base import DeflectorsBase as DeflectorsBase
--- a/slsim/Deflectors/__init__.py
+++ b/slsim/Deflectors/__init__.py
@@ -1,13 +1,13 @@
-from .all_lens_galaxies import *
-from .compound_lens_halos_galaxies import *
-from .elliptical_lens_galaxies import *
-from .deflector import *
-from .deflectors_base import *
+from .all_lens_galaxies import AllLensGalaxies
+from .compound_lens_halos_galaxies import CompoundLensHalosGalaxies
+from .elliptical_lens_galaxies import EllipticalLensGalaxies
+from .deflector import Deflector
+from .deflectors_base import DeflectorsBase
 
-# from .all_lens_galaxies import AllLensGalaxies as AllLensGalaxies
-# from .compound_lens_halos_galaxies import (
-#     CompoundLensHalosGalaxies as CompoundLensHalosGalaxies,
-# )
-# from .elliptical_lens_galaxies import EllipticalLensGalaxies as EllipticalLensGalaxies
-# from .deflector import Deflector as Deflector
-# from .deflectors_base import DeflectorsBase as DeflectorsBase
+__all__ = [
+    "AllLensGalaxies",
+    "CompoundLensHalosGalaxies",
+    "EllipticalLensGalaxies",
+    "Deflector",
+    "DeflectorsBase",
+]
--- a/slsim/Deflectors/__init__.py
+++ b/slsim/Deflectors/__init__.py
@@ -1,0 +1,5 @@
+from .all_lens_galaxies import AllLensGalaxies
+from .compound_lens_halos_galaxies import CompoundLensHalosGalaxies
+from .elliptical_lens_galaxies import EllipticalLensGalaxies
+from .deflector import Deflector
+from .deflectors_base import DeflectorsBase
--- a/slsim/Sources/QuasarCatalog/__init__.py
+++ b/slsim/Sources/QuasarCatalog/__init__.py
@@ -1,1 +1,3 @@
-from .quasar_plus_galaxies import quasar_galaxies_simple as quasar_galaxies_simple
+from .quasar_plus_galaxies import quasar_galaxies_simple
+
+__all__ = ["quasar_galaxies_simple"]
--- a/slsim/Sources/QuasarCatalog/__init__.py
+++ b/slsim/Sources/QuasarCatalog/__init__.py
@@ -1,0 +1,1 @@
+from .quasar_plus_galaxies import quasar_galaxies_simple as quasar_galaxies_simple
--- a/slsim/Sources/SupernovaeCatalog/__init__.py
+++ b/slsim/Sources/SupernovaeCatalog/__init__.py
@@ -1,0 +1,1 @@
+from .supernovae_sample import SupernovaeCatalog as SupernovaeCatalog
--- a/slsim/Sources/SupernovaeCatalog/__init__.py
+++ b/slsim/Sources/SupernovaeCatalog/__init__.py
@@ -1,1 +1,3 @@
-from .supernovae_sample import SupernovaeCatalog as SupernovaeCatalog
+from .supernovae_sample import SupernovaeCatalog
+
+__all__ = ["SupernovaeCatalog"]
--- a/slsim/Sources/__init__.py
+++ b/slsim/Sources/__init__.py
@@ -4,3 +4,4 @@ from .point_plus_extended_sources import (
     PointPlusExtendedSources as PointPlusExtendedSources,
 )
 from . import QuasarCatalog as QuasarCatalog
+from . import SupernovaeCatalog as SupernovaeCatalog
--- a/slsim/Sources/__init__.py
+++ b/slsim/Sources/__init__.py
@@ -1,0 +1,3 @@
+from .galaxies import Galaxies
+from .point_sources import PointSources
+from .point_plus_extended_sources import PointPlusExtendedSources
--- a/slsim/Sources/__init__.py
+++ b/slsim/Sources/__init__.py
@@ -3,3 +3,4 @@ from .point_sources import PointSources as PointSources
 from .point_plus_extended_sources import (
     PointPlusExtendedSources as PointPlusExtendedSources,
 )
+import QuasarCatalog as QuasarCatalog
--- a/slsim/Sources/__init__.py
+++ b/slsim/Sources/__init__.py
@@ -1,3 +1,5 @@
-from .galaxies import Galaxies
-from .point_sources import PointSources
-from .point_plus_extended_sources import PointPlusExtendedSources
+from .galaxies import Galaxies as Galaxies
+from .point_sources import PointSources as PointSources
+from .point_plus_extended_sources import (
+    PointPlusExtendedSources as PointPlusExtendedSources,
+)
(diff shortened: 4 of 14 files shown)


slsim/Deflectors/__init__.py
```diff
@@ -0,0 +1,7 @@
+from .all_lens_galaxies import AllLensGalaxies as AllLensGalaxies
```
sibirrer, line 1: just asking, is there a particular reason to put these import statements in the init file? This may be harder to maintain in the development to make sure the relevant classes and definitions are important.

slsim/lens_pop.py
sibirrer, line 1: I don't see the new LensPop class anywhere

sibirrer (changes requested):
Thank you for the PR @jacob-hjortlund !
I have two questions on the PR. Then there is a minor test failure. I hope you can test it locally and fix it

slsim/lens_pop.py
jacob-hjortlund: No new LensPop class in this PR, only moving pipelines, deflector population and source population objects outside of the class. Additional refactors will come down the line once other components are completed, but currently tackling this in steps to maintain compatibility.

sibirrer: It is just marked as deleted here, or I can't see the code

slsim/Deflectors/__init__.py
jacob-hjortlund: My main reasoning is cleaning up imports across SLSim. The previous implementation of LensPop had multiple imports from slsim.Deflectors and slsim.Sources, which can be cleaned up by treating these as submodules with relevant imports brought in via their respective __init__.py files.

I'm not sure I follow why this would be harder to maintain, unless you're referring to the re-exports and potential future name changes of classes within these submodules?

slsim/lens_pop.py
```diff
@@ -1,378 +1,70 @@
...
+
         super().__init__(
-            sky_area,
-            cosmo,
-            lightcurve_time,
-            sn_type,
-            sn_absolute_mag_band,
-            sn_absolute_zpsys,
-            sn_modeldir,
+            sky_area=None,
```
sibirrer, line 50: there should be a sky_area quantity in LensPop. This should be particularly different then the sky area of the deflector and/or source population to make it easier to create larger sky area realizations

jacob-hjortlund:
> Thank you for the PR @jacob-hjortlund ! I have two questions on the PR. Then there is a minor test failure. I hope you can test it locally and fix it

Was this minor test failure before [9f4bd72](https://github.com/LSST-strong-lensing/slsim/pull/236/commits/9f4bd72f9bc1e7d2184057d1d4f8b341b928a715)? Currently seeing all tests passing on my end.

slsim/lens_pop.py
```diff
@@ -1,378 +1,70 @@
...
-        kwargs_quasars_galaxies=None,
-        variability_model=None,
-        kwargs_variability=None,
-        kwargs_mass2light=None,
-        skypy_config=None,
-        slhammocks_config=None,
-        sky_area=None,
-        source_sky_area=None,
-        deflector_sky_area=None,
-        filters=None,
```
sibirrer, line 33: how do we make sure that the correct filters are chosen in the Deflector and Source classes? Could cause issues if they are not the same.

slsim/lens_pop.py
```diff
@@ -1,378 +1,70 @@
...
-        :param los_config: configuration for line of sight distribution
-        :type los_config: LOSConfig instance
-        :param sn_modeldir: sn_modeldir is the path to the directory containing files
-         needed to initialize the sncosmo.model class. For example,
-         sn_modeldir = 'C:/Users/username/Documents/SALT3.NIR_WAVEEXT'. These data can
-         be downloaded from https://github.com/LSST-strong-lensing/data_public .
-         For more detail, please look at the documentation of RandomizedSupernovae
-         class.
-        :type sn_modeldir: str
+        Args:
```
sibirrer, line 30: can you follow the :param input_name: and :type input_name: conventions here?

slsim/lens_pop.py
```diff
@@ -1,378 +1,70 @@
...
-                self._source_model_type = "point_plus_extended"
-        else:
-            raise ValueError("source_type %s is not supported" % source_type)
+        self._lens_galaxies = deflector_population
+        self._sources = source_population
 
         self._factor_source = self.f_sky.to_value(
             "deg2"
-        ) / self.source_sky_area.to_value("deg2")
+        ) / self._sources._sky_area.to_value("deg2")
```
sibirrer, line 64: we should avoid accessing private variables of other classes. Use self._sources.sky_area instead

slsim/lens_pop.py
```diff
@@ -434,7 +126,7 @@ def get_num_sources_tested_mean(self, testarea):
         """
         num_sources = self.source_number
         num_sources_tested_mean = (testarea * num_sources) / (
-            12960000 * self._factor_source * self.source_sky_area.to_value("deg2")
+            12960000 * self._factor_source * self._sources._sky_area.to_value("deg2")
```
sibirrer, line 129: no access to private variable in other classes. Make _sky_area a public variable as sources.sky_area

sibirrer (changes requested):
some more minor comments and requests

slsim/Deflectors/__init__.py
sibirrer: Yes, I was meaning for further potential name changes or addition of classes, these init files would also need to be updated

slsim/lens_pop.py
jacob-hjortlund: How is this the case in the current LensPop implementation? As far as I understand, if we provide a deflector and/or source population sky_area, the LensPop sky_area is ignored. Have I misunderstood this?

slsim/lens_pop.py
jacob-hjortlund: Good catch! I'd add a ValueError in LensPop to catch this, since this is a requirement from LensPop and all source / deflector related attributes should be owned by their respective classes.

slsim/lens_pop.py
sibirrer: The LensPop sky_area is the ultimate area over which strong lensing populations are simulated. If the source and/or deflector sky areas are different, then the multiplicative factor between the final sky area and the sub-population is computed and then the individual deflector/source populations are up (or down) sampled from.
I.e. we do not need 100 billion of galaxies (half the sky or so) to simulate lenses on half the sky. As long as we have a much smaller but still representative area, we can just re-use the much smaller populations multiple times (randomly)

jacob-hjortlund: Thanks, helped me realise I'd missed the f_sky in the lens / deflector factors. Makes sense, will fix asap.

slsim/lens_pop.py
jacob-hjortlund: Definitely can. Just curious, is there a specific reason we use the harder-to-read rEST format vs. the easier-to-read Google format, considering they are both supported by sphinx? One could argue that better docstring readability in the source code could promote contribution. :smile:

sibirrer: no, this was mainly the initial setting and the current sphinx configuration. If you want to change the sphinx configuration, that would work, just would eventually need to update many docstrings, that's all

slsim/Deflectors/__init__.py
jacob-hjortlund: Currently we have multiple imports from within a single submodule. A rename of a class would require changes in the submodule as well as for any files that import the renamed class. If we instead make use of __init__.py, we have to make a similar amount of changes (with one additional change being in the submodule __init__.py), with the benefit of having a clear and readable import structure. A similar argument holds for addition of new classes or functions that we wish to be public facing.

I'm not sure I see the point in organizing things within submodules, if we do not make use of the __init__.py file while needing to access multiple files within said submodule.

slsim/lens_pop.py
jacob-hjortlund: If you're okay with that , then I'd love to push for it! Can add it as an issue for the refactor. :star_struck:

sibirrer: I am ok with it :)

slsim/Sources/QuasarCatalog/__init__.py
```diff
@@ -0,0 +1 @@
+from .quasar_plus_galaxies import quasar_galaxies_simple as quasar_galaxies_simple
```
nkhadka21, line 1: Is it neceassary to use alias?

slsim/Sources/SupernovaeCatalog/__init__.py
```diff
@@ -0,0 +1 @@
+from .supernovae_sample import SupernovaeCatalog as SupernovaeCatalog
```
nkhadka21, line 1: Here also, Is it neceassary to use alias?

slsim/Sources/__init__.py
```diff
@@ -0,0 +1,13 @@
+from .galaxies import Galaxies
```
nkhadka21, line 1: Does importing all of these at the package level slow the import time in some cases?

nkhadka21 (commented):
Hi @jacob-hjortlund, thank you very much for this PR! I have some minor comments. Please look at them.

slsim/Sources/__init__.py
jacob-hjortlund: I wouldn't expect this to slow things considerably and doubt this is the biggest runtime bottleneck in the code, but can run a profiler with and without submodule imports if you'd prefer?

slsim/Sources/QuasarCatalog/__init__.py
jacob-hjortlund: Either aliasing or specifiying in __all__ for the submodule is necessary to uphold linter requirements from ruff. Missed these when switching to __all__ instead of aliasing, thanks for the catch!

slsim/Sources/__init__.py
jacob-hjortlund: Either aliasing or specifiying in __all__ for the submodule is necessary to uphold linter requirements from ruff. Missed these when switching to __all__ instead of aliasing, thanks for the catch!

slsim/Sources/SupernovaeCatalog/__init__.py
jacob-hjortlund: Either aliasing or specifiying in __all__ for the submodule is necessary to uphold linter requirements from ruff. Missed these when switching to __all__ instead of aliasing, thanks for the catch!

slsim/lens_pop.py
jacob-hjortlund: Coming back to this, how willing are we to let this be up to the user? Currently the filters kwarg is only used in the case of galaxy-galaxy strong lensing generated via SkyPyPipeline. For other sources the user has to make sure that kwargs_source_cut selects the same filters as provided by the deflector pipeline.

I'd argue for leaving this be in the current PR and adding it as a task for #225 , where we can make the available filters an attribute, such that we can add a check in LensPop as mentioned in my previous comment.

slsim/Sources/__init__.py
nkhadka21: It would be good to check!

slsim/lens_pop.py
sibirrer: Hi @jacob-hjortlund the kwargs_source_cut is only when you want to make cuts in the distribution. SNe also need to be provided filters (and so do quasars). @nkhadka21 made an issue in the past to homogenize the filter naming among the different galaxy/point-source classes. Perhaps you can leave it as a #TODO in the initialization, or define a routine check_consistency() in LensPop that is generally operating on class instances of sources and deflectors and might help diagnose issues for the user to be raised

slsim/Sources/__init__.py
jacob-hjortlund: I used the python importtime profiler + tuna for visualization to profile a simple script which only contains imports. In the case of original.log, this script contains all imports in LensPop for the case of galaxy-galaxy strong lensing, while refactor.log contains the imports in this PR's lens_pop.py.

You can visualize the logs by running `pip install tuna` and `tuna original.log`. This simple profiling run shows that the original setup takes 2.301 s while the refactored imports take 1.571 s, including overhead from the profiler itself.

[original.log](https://github.com/user-attachments/files/16638199/original.log)
[refactor.log](https://github.com/user-attachments/files/16638200/refactor.log)

sibirrer: Thank you very much @jacob-hjortlund ! This is very good to know!

slsim/lens_pop.py
jacob-hjortlund: In that case I'll add this as a TODO in the init, since source classes will be modified in #225 and onwards.

sibirrer (approved):
Looks good to me. Thank you @jacob-hjortlund !

jacob-hjortlund:
@sibirrer great to hear. @sibirrer @nkhadka21 I don't have write access, could one of you smack the big green button?

sibirrer:
Pinging here @nkhadka21 for final review and potential merge

nkhadka21 (changes requested):
Hi @jacob-hjortlund, thank you very much for this wonderful work! It looks very good. Currently, users need to initialize some of the classes before they can initialize the LensPop class, which is very good. It would also be great to have an API somewhere in Util that can initialize all these classes in one place for users who want to minimize effort. Having this option could be very useful.

sibirrer:
> Hi @jacob-hjortlund, thank you very much for this wonderful work! It looks very good. Currently, users need to initialize some of the classes before they can initialize the LensPop class, which is very good. It would also be great to have an API somewhere in Util that can initialize all these classes in one place for users who want to minimize effort. Having this option could be very useful.

@nkhadka21 perhaps the API can be made once the total refactoring is done as the to be initialized classes might also change down the road?

nkhadka21:
> > Hi @jacob-hjortlund, thank you very much for this wonderful work! It looks very good. Currently, users need to initialize some of the classes before they can initialize the LensPop class, which is very good. It would also be great to have an API somewhere in Util that can initialize all these classes in one place for users who want to minimize effort. Having this option could be very useful.
> 
> @nkhadka21 perhaps the API can be made once the total refactoring is done as the to be initialized classes might also change down the roa

nkhadka21 (approved):

PR #239, opened 2024-08-20
make it possible to insert directly specific kappa and gamma values into a Lens() object

sibirrer:


nkhadka21 (approved):
Hi @sibirrer, thank you for this PR! Looks good to me.

PR #241, opened 2024-08-21
Allows for initialization of Source or Lens class using dicts instead of astropy tables

ahuang314:
Previously, in order to create an instance of the Lens class, the source_dict parameter needed to be an astropy table (the documentation was also incorrect as a consequence). This PR allows for source_dict to be a dict by converting the dict into an astropy table during the initialization of the class.

Additionally, if the user wants to initialize a Lens or Source class with a specific center_x and center_y, they can now do so by specifying them in source_dict.

### DIFF
--- a/slsim/Sources/source.py
+++ b/slsim/Sources/source.py
@@ -56,13 +56,18 @@ class Source(object):
         # Convert dict to astropy table
         if isinstance(source_dict, dict):
             self.source_dict = Table([source_dict])[0]
-        else: #if source_dict is already an astropy table
+        else:  # if source_dict is already an astropy table
             self.source_dict = source_dict
 
         # If center_x and center_y are already specified, use them instead of picking randomly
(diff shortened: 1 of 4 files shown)


sibirrer (approved):
Thank you @ahuang314 ! Looks good to me. I will leave it to @nkhadka21 for perhaps further review. I also link @jacob-hjortlund here, as he might face similar issues and just keep him in the loop during his refactoring

nkhadka21 (commented):
Hi @ahuang314 , thank you very much for this PR! It looks good to me. I noticed that testing time increased from 22 min to ~30 min. Can you check what is causing this and try to reduced this time.

ahuang314:
@nkhadka21 I am not sure why it takes 30 minutes now. Actually, it took 30 minutes in PR #236 as well. I will try rerunning the tests and see if it is consistently slow

PR #242, opened 2024-08-21
Deflector and Source angular size unit consistency.

nkhadka21:
This PR maintains consistency for the source and deflector angular size unit.


sibirrer (approved):
Looks good, thank you @nkhadka21 !

PR #243, opened 2024-08-22
Added functionality to slsim to use galsim+webbpsf to generate Roman images

ahuang314:
Summary of changes:

1. New pipeline for simulating Roman images in `roman_image_simulation.py`
2. Slight modification to `image_simulation.py` which allows for the user to overwrite default values in kwargs_band and kwargs_numerics
3. Slight modification to the roman_configurations notebook in order to account for the refactoring of LensPop, as well as showing off the new `simulate_roman_images` function

Not too much testing is done for `simulate_roman_images`; since webbpsf requires external files, we cannot test the new functions entirely and we have to trust that webbpsf and galsim work properly.


nkhadka21 (changes requested):
Hi @ahuang314, thank you very much for this PR! It looks very good to me. I have one comment. Please look at it. This is also a question for @sibirrer!

slsim/roman_image_simulation.py
```diff
@@ -0,0 +1,331 @@
...
+from lenstronomy.SimulationAPI.sim_api import SimAPI
+from slsim.Observations import image_quality_lenstronomy
+import os.path
+import pickle
+from webbpsf.roman import WFI
+
+# NOTE: Adding sky background requires webbpsf-data, which can be found at
+#       https://webbpsf.readthedocs.io/en/latest/installation.html. Then, the
+#       environment variable "WEBBPSF_PATH" must be set to this webbpsf-data directory.
+#       Additionally, the galsim module is required, which is not supported on Windows.
```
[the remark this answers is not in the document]
sibirrer: yes, having galsim only in the test requirements is better, and then in the import of galsim, we can do a try: import galsim
except: raiseImportError("please install Galsim")
or something along these lines

requirements.txt
```diff
@@ -11,3 +11,4 @@ sncosmo
 colossus
 uncertainties
 kcorrect
+webbpsf
```
sibirrer, line 14: just checking, this is easy installable and being used, or otherwise could also be moved to the test_requirements?

slsim/Plots/lens_plots.py
```diff
@@ -36,31 +39,70 @@ def rgb_image(self, lens_class, rgb_band_list, add_noise=True):
...
+                num_pix=self.num_pix,
+                oversample=5,
+                add_noise=add_noise,
+                **self._kwargs
+            )
+            image_b = simulate_roman_image(
+                lens_class,
+                band=rgb_band_list[2],
+                num_pix=self.num_pix,
+                oversample=5,
```
sibirrer, line 64: here you use a oversample=5 argument. Is this differently treated for the other observatories? Maybe we can uniformly treat it such that it could also be used for others? Perhaps with the argument kwargs_numerics?

slsim/Plots/lens_plots.py
```diff
@@ -36,31 +39,70 @@ def rgb_image(self, lens_class, rgb_band_list, add_noise=True):
...
-        image_b = simulate_image(
-            lens_class=lens_class,
-            band=rgb_band_list[2],
-            num_pix=self.num_pix,
-            add_noise=add_noise,
-            observatory=self._observatory,
-            **self._kwargs
-        )
-        image_rgb = make_lupton_rgb(image_r, image_g, image_b, stretch=0.5)
+        if self._observatory == "Roman":
```
sibirrer, line 42: maybe all these inputs can be shortened if the simulate_roman_image() and simulate_image() follow exactly the same API? Then we only have to overwrite/define the function to be called as an if/else statement and everything else is identical

slsim/roman_image_simulation.py
```diff
@@ -0,0 +1,315 @@
...
+    else:
+        psf_file_path = os.path.join(
+            os.path.dirname(__file__), "..", "data", "webbpsf", psf_file_name
+        )
+
+    if os.path.exists(psf_file_path):
+        with open(psf_file_path, "rb") as psf_file:
+            psf = pickle.load(psf_file)
+    else:
+        wfi = WFI()
```
sibirrer, line 201: is it possible to test these lines of code?

sibirrer (commented):
Thank you very much @ahuang314! I have few comments/questions for minor improvements.

slsim/roman_image_simulation.py
ahuang314: No, these lines of code are the ones that require external files from webbpsf to generate the psf

requirements.txt
ahuang314: webbpsf does not have restrictions on operating system so there is no need to move it to test_requirements. It can also be imported with no problems. The only thing is that in order to actually use the functions and classes within the module, external files are needed. If the user tries to use webbpsf features without these external files, webbpsf has its own very beautiful error message that links the user to the files.

slsim/Plots/lens_plots.py
ahuang314: I can try to shorten this part of the code

slsim/Plots/lens_plots.py
ahuang314: Yes, in `simulate_roman_image`, the oversample is outside of `kwargs_numerics` because we want `image_model.image` to return the still oversampled image, and we do that by making an instance of the `SimAPI` class with `numpix = num_pix * oversample` and `pixel_scale = pixel_scale / oversample`. This is because we do the psf convolution using `galsim` instead of inside the `image_model` class.

It is still possible to put oversample in `kwargs_numerics` but then we have to do some gymnastics of extracting the oversample value and then changing it back to 1 after creating an instance of the SimAPI class so then it does not oversample again when we use `image_model.image`. I think it is better to just have the oversample parameter explicitly in the function arguments

sibirrer (approved):
Thank you @ahuang314 ! I let @nkhadka21 do the further review and merge

nkhadka21 (commented):
Hi @ahuang314 , thank you very much for this quick update! looks very good. I have a minor comment. The file in data/webbpsf is 855kb. It would be great if you could reduce the size of this file. You might try a single psf in all cases.

ahuang314:
@nkhadka21 The 855kb is already only one single psf

ahuang314:
It's 405 kb now

nkhadka21:
> It's 152 kb now

Much better! Thank you @ahuang314.

nkhadka21 (approved):

PR #250, opened 2024-09-10
Debug for LensPop got same LOS

trivialTZ:
1. add LOSConfigParams so when LensPop rendering population, will got different LOSConfig class.
2. delete the invalid input of mixgauss_means....


trivialTZ:
Minor debug update for Line of Sight used in LensPop. For the codecov of the code, I believe all the change of the code were aleady be tested.
@sibirrer @nkhadka21

sibirrer (changes requested):
Thank you @trivialTZ ! I have one minor request for the testing.

PR #253, opened 2024-09-18
Notebook updates

nkhadka21:
This PR updates some of the existing notebooks. Additionally, it has applied a correction factor to the velocity dispersion function.


slsim/Deflectors/velocity_dispersion.py
```diff
@@ -353,7 +353,10 @@ def schechter_vel_disp(
         resolution=1000,
         scale=1,
     )
-    return z, vel_disp
+    # above vel_disp is measured with in the effective radius/8. So, correction factor
+    # of 8**(-0.066) must be applied to convert to vel_disp at effective radius.
```
sibirrer, line 357: can you provide a reference or more detailed calculation for this?

sibirrer (changes requested):
Looks good @nkhadka21 I have one minor request

slsim/Deflectors/velocity_dispersion.py
nkhadka21: Added a reference for this correction factor.

sibirrer (approved):
Thank you @nkhadka21. Please merge after passing of the CIs

PR #255, opened 2024-09-19
change convention of shear to be consistent with SLSim conventions of ellipticity

sibirrer:


nkhadka21 (approved):
Thanks @sibirrer for fixing this!

PR #260, opened 2024-10-08
minor updates in slsim coolest interface.

nkhadka21:
A minor update in slsim coolest interface.

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -11,4 +11,4 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - uses: actions/setup-python@v3
-    - uses: pre-commit/action@v3.0.1
+    - uses: pre-commit/action@v3.0.2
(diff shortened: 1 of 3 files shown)


sibirrer (commented):
Thank you @nkhadka21! Do you know why this was not spotted beforehand, or did coolest update since? Then the pre-commit CI is not working, so perhaps we should also fix it with this commit (perhaps down-grade it?)

nkhadka21:
> Thank you @nkhadka21! Do you know why this was not spotted beforehand, or did coolest update since? Then the pre-commit CI is not working, so perhaps we should also fix it with this commit (perhaps down-grade it?)

It was because of the change in python version. It works for python3.8 and this error comes only when we use python3.9. It seems like pre-commit has some issues and hope they will fix it from their side. I will also try to solve it if it is possible.

nkhadka21:
They are saying, one should set pre-commit<4. We are using v3. I think we should ignore this issue for now.

sibirrer (approved):

PR #262, opened 2024-10-11
False positive generation

nkhadka21:
This PR implements a code to generate false positive for the galaxy galaxy lenses.


slsim/false_positive.py
```diff
@@ -0,0 +1,314 @@
+import numpy as np
+from lenstronomy.Cosmo.lens_cosmo import LensCosmo
+from slsim.ParamDistributions.los_config import LOSConfig
+from slsim.Util.param_util import ellipticity_slsim_to_lenstronomy
+from slsim.lens import theta_e_when_source_infinity
+
+
+class FalsePositive(object):
```
sibirrer, line 8: there is a lot of duplicated code here with the Lens() class. Could you inherit the Lens() class and then only alter the definitions that really are different?

slsim/false_positive_pop.py
```diff
@@ -0,0 +1,111 @@
+from slsim.false_positive import FalsePositive
```
sibirrer, line 1: maybe also move this file to a separate folder away from the base of SLSim

slsim/false_positive_pop.py
```diff
@@ -0,0 +1,111 @@
...
+        :param number: number of false positive requested. The default value is 1.
+        :return: list of FalsePositive() instance with parameters of a pair of
+            elliptical and blue galaxy.
+        """
+        false_positive_population = []
+
+        for _ in range(number):
+            successful = False
+            while not successful:
+                try:
```
sibirrer, line 59: try - except is very dangerous because you might not know from where the error came. Can you instead make conditions instead of errors breaking the code?

slsim/lens.py
```diff
@@ -670,18 +673,15 @@ def source_light_model_lenstronomy(self, band=None):
             self._source_type == "extended"
             or self._source_type == "point_plus_extended"
         ):
-            if self.light_profile == "single_sersic":
+            if self.source.light_profile == "single_sersic":
```
sibirrer, line 676: is it perhaps possible to make this entire definition in the Source() class and providing the necessary inputs from the Lens() class?

slsim/lens_pop.py
```diff
@@ -71,19 +71,25 @@ def __init__(
...
 
-    def select_lens_at_random(self, **kwargs_lens_cut):
+    def select_lens_at_random(self, test_area=None, **kwargs_lens_cut):
         """Draw a random lens within the cuts of the lens and source, with possible
         additional cut in the lensing configuration.
 
         # TODO: make sure mass function is preserved, # as well as option to draw all
         lenses within the cuts within the area
 
+        :param test_area: area of disk around one lensing galaxies to be investigated on
```
sibirrer, line 81: "disk", I'd rather say "solid angle", just a minor comment

sibirrer (changes requested):
Thank you very much @nkhadka21 . I have some comments and request on the design. Let me know if you have questions

slsim/false_positive.py
nkhadka21: Hi @Simon, Lens() class and FalsePositive() class has different inputs. Also, FalsePositive() class takes multiple sources while Lens class has only single source option. These difference create some conflicts while inheriting Lens class in FalsePositive class. My plan is to refactor Lens class first and then inherit in falsePositive class. But, this will make this PR very large. So, can you do this in next PR?

sibirrer: @nkhadka21 (first, my tag is @sibirrer and not <at>simon (which notified someone totally unrelated)). How urgently does this PR need to be merged? The reason I am asking is also that we could refactor in parallel the Lens() class in a different PR

nkhadka21: Thank you @sibirrer for correcting this! I think it not very urgent to merge this PR. I agree that we can refactor Lens() class first and then update this PR.

PR #266, opened 2024-10-22
excluded LsstSciencePipeline from codecov

nkhadka21:


sibirrer (commented):
can we have the definition opsim_variable_lens_injection() being moved somewhere where it gets tested?

PR #268, opened 2024-10-25
Different option for power law slope

nkhadka21:
This PR implements different options for the power law slope.

### DIFF
--- a/slsim/Deflectors/elliptical_lens_galaxies.py
+++ b/slsim/Deflectors/elliptical_lens_galaxies.py
@@ -16,6 +16,7 @@ class EllipticalLensGalaxies(DeflectorsBase):
         kwargs_mass2light,
         cosmo,
         sky_area,
+        gamma_pl=None,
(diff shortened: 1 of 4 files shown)


slsim/Deflectors/elliptical_lens_galaxies.py
```diff
@@ -59,7 +65,31 @@ def __init__(
         if "n_sersic" not in column_names:
             galaxy_list["n_sersic"] = -np.ones(n)
         if "gamma_pl" not in column_names:
-            galaxy_list["gamma_pl"] = np.ones(n) * 2
+            if gamma_pl is None:
```
sibirrer, line 68: is there a way where we perhaps do not have to duplicate the code in two classes but instead use one routine/definition for it?

sibirrer (changes requested):
Looks good. I just think if there is a way to avoid the duplication of the same code for the two classes but instead inherit from a same base class?

PR #269, opened 2024-10-25
Lens class refactor

nkhadka21:
This PR refactors the Lens Class and implements multi-source lens.


slsim/Sources/galaxies.py
```diff
@@ -120,14 +120,24 @@ def source_number_selected(self):
         """
         return self._num_select
 
-    def draw_source(self):
-        """Choose source at random.
+    def draw_source(self, z_max=None):
+        """Choose source at random. :param z_max: maximum redshift for source to be
```
sibirrer, line 124: doc strings a bit messed up

slsim/Sources/source.py
```diff
@@ -483,15 +493,15 @@ def point_source_position(self, center_lens, draw_area):
...
         return extended_source_center
 
-    def kwargs_extended_source_light(
-        self, center_lens, draw_area, band=None, light_profile_str="single_sersic"
-    ):
+    def kwargs_extended_source_light(self, center_lens, draw_area, band=None):
         """Provides dictionary of keywords for the source light model(s). Kewords used
         are in lenstronomy conventions.
 
+        :param center_lens: center of the deflector.
```
sibirrer, line 500: do you want to add the units of arc seconds?

slsim/Sources/source.py
```diff
@@ -483,15 +493,15 @@ def point_source_position(self, center_lens, draw_area):
...
+         Eg: np.array([center_x_lens, center_y_lens])
+        :param draw_area: The area of the test region from which we randomly draw a
```
sibirrer, line 502: do you want to add that this is a solid angle in units of square arc seconds?

slsim/lens.py
sibirrer, line 1: just checking, don't you want to instantiate Lens() with already instantiated Source() and Deflector() class objects? I guess this is not complete yet. Just ping me when it's ready :)

sibirrer (commented):
Just some initial comments

slsim/lens.py
nkhadka21: Yes, you are right. I had created this PR without commiting the changes.

slsim/lens.py
```diff
@@ -320,27 +277,36 @@ def deflector_redshift(self):
...
     @property
     def external_shear(self):
         """
 
         :return: the absolute external shear
         """
-        gamma1, gamma2, _ = self.los_linear_distortions
-        return (gamma1**2 + gamma2**2) ** 0.5
+        shear_list = []
+        for los_linear_dist in self.los_linear_distortions:
```
sibirrer, line 306: I am not sure I understand the looping through here. There is only one line of sight per lens. The only thing that changes is the lensing efficiency part

slsim/lens.py
```diff
@@ -320,27 +277,36 @@ def deflector_redshift(self):
...
     @property
     def external_convergence(self):
         """
 
         :return: external convergence
         """
-        _, _, kappa_ext = self.los_linear_distortions
-        return kappa_ext
+        kappa_list = []
+        for los_linear_dist in self.los_linear_distortions:
```
sibirrer, line 294: I am not sure I understand the looping through here. There is only one line of sight per lens. The only thing that changes is the lensing efficiency part

slsim/lens.py
```diff
@@ -349,33 +315,43 @@ def einstein_radius_deflector(self):
 
         :return:
```
sibirrer, line 316: can you specify that this is now a list that is returned?

slsim/lens.py
```diff
@@ -409,14 +385,18 @@ def los_linear_distortions(self):
         return self._los_linear_distortions_cache
 
     def _calculate_los_linear_distortions(self):
-        """Line-of-sight distortions in shear and convergence.
+        """Line-of-sight distortions in shear and convergence for each source.
 
         :return: kappa, gamma1, gamma2
```
sibirrer, line 390: do you want to specify that these are lists now?

slsim/lens.py
```diff
@@ -430,23 +410,27 @@ def deflector_magnitude(self, band):
...
+        images arriving later. This is for single source.
+
+        :return: list of arrival times for each image [days] for each source.
+        :rtype: list of numpy array
+        """
+        lens_model_list, kwargs_lens =  self.deflector_mass_model_lenstronomy()
+        image_pos = self.point_source_image_positions()
+        arrival_times_list = []
+        for index, source in enumerate(self.source):
+            lens_model = LensModel(
```
sibirrer, line 422: just fyi, in the latest version of lenstronomy (not yet pip version), it supports arrival times with multiple source redshifts. No need to do it right now

slsim/lens.py
```diff
@@ -460,20 +444,24 @@ def image_observer_times(self, t_obs):
...
-                t_obs[:, np.newaxis] - arrival_times + np.min(arrival_times)
-            ).T
-        else:
-            observer_times = (t_obs - arrival_times + np.min(arrival_times))[
-                :, np.newaxis
-            ]
-
-        return observer_times
+        observer_times_list = []
+        for point_source_arrival_time in self.point_source_arrival_times():
```
sibirrer, line 448: just making sure: this routine is not yet implemented for multiple source redshift, hence this function only works in limited ways. Perhaps add a TODO or warning to make sure users are not (yet) using it for multiple source planes

slsim/lens.py
```diff
@@ -517,69 +514,91 @@ def extended_source_magnitude(self, band, lensed=False):
...
-                self._extended_source_magnification = flux_lensed / flux_no_lens
-            else:
-                self._extended_source_magnification = 0
-        return self._extended_source_magnification
+                num_pix = 200
+                delta_pix = theta_E * 4 / num_pix
+                x, y = util.make_grid(numPix=num_pix, deltapix=delta_pix)
+                x += center_source[0]
+                y += center_source[1]
+                beta_x, beta_y = lensModel.ray_shooting(x, y, kwargs_params["kwargs_lens"])
```
sibirrer, line 588: the lensModel() object needs to know the source redshift to do the ray-tracing, so that does not work. I need to make a new lenstronomy release (it's implemented but not on pip), then we can change the source redshift such that the ray_shooting() knows what source redshift to pick

slsim/lens.py
```diff
@@ -593,11 +612,19 @@ def lenstronomy_kwargs(self, band=None):
...
         kwargs_model = {
             "lens_light_model_list": lens_light_model_list,
             "lens_model_list": lens_mass_model_list,
         }
+        if self.source_number > 1:
+            kwargs_model["lens_redshift_list"] = [
+                self.deflector_redshift]*len(lens_mass_model_list)
+            kwargs_model["z_lens"] = self.deflector_redshift
+            kwargs_model["z_source"] = self.single_source_class.redshift
+            kwargs_model["source_redshift_list"] = self.source_redshift
```
sibirrer, line 625: do you want to rename self.source_redshift since it's a list here

slsim/lens.py
```diff
@@ -670,19 +698,25 @@ def source_light_model_lenstronomy(self, band=None):
             self._source_type == "extended"
             or self._source_type == "point_plus_extended"
         ):
-            if self.light_profile == "single_sersic":
-                source_models["source_light_model_list"] = ["SERSIC_ELLIPSE"]
+            source_models_list = []
+            kwargs_source_list = []
+            for source in self.source:
```
sibirrer, line 703: I would re-structure this since we might want different source types for different sources supported

slsim/lens.py
```diff
@@ -670,19 +698,25 @@ def source_light_model_lenstronomy(self, band=None):
...
+                source_models_list.append(source.extended_source_light_model())
```
sibirrer, line 704: for the different source redshifts, I think we need to save also the source_redshift_list as well, otherwise the danger is that people are using this setting and forget to also query the source redshift list

slsim/lens_pop.py
```diff
@@ -84,28 +86,35 @@ def select_lens_at_random(self, test_area=None, **kwargs_lens_cut):
         while True:
```
sibirrer, line 86: write that this is a single lens single source lens that is created

slsim/lens_pop.py
```diff
@@ -178,41 +187,57 @@ def draw_population(self, kwargs_lens_cuts, speed_factor=1):
...
-                        gg_lens_population.append(gg_lens)
-                        # if a lens system passes the validity test, code should exit
-                        # the loop. so, n should be greater or equal to
-                        # num_sources_tested which will break the while loop
-                        # (instead of this one can simply use break).
-                        n = num_sources_tested
+                    if lens_class.validity_test(**kwargs_lens_cuts):
+                        valid_sources.append(_source)
+                    n += 1
+                if valid_sources:
```
sibirrer, line 225: this is dangerous. Perhaps make the conditions explicit, i.e. if len(valid_sources) > 0:

slsim/lens.py
nkhadka21: Yes, it is not necessary to loop through all source. They result in the same value. I was confused by the input of the function!

slsim/lens.py
nkhadka21: I changed it back to original one.

slsim/lens.py
nkhadka21: There are more complex issues when we try to allow difference source type. So, I think it is good to work on this with a separate PR.

slsim/lens.py
```diff
@@ -341,41 +311,53 @@ def external_shear(self):
...
+                                    cosmo=self.cosmo,
+                                )
+                    _theta_E = _lens_cosmo.sis_sigma_v2theta_E(
+                        float(self.deflector.velocity_dispersion(cosmo=self.cosmo))
+                    )
+                    self._theta_E_list.append(_theta_E)
+                else:
+                    # numerical solution for the Einstein radius
+                    lens_model_list, kwargs_lens = self.deflector_mass_model_lenstronomy()
+                    lens_model = LensModel(lens_model_list=lens_model_list)
```
sibirrer, line 340: for different source redshifts, this does not work. This way it only works for z_source=z_source_convention. I'll update the lenstronomy version such that we can adopt to it

slsim/lens.py
```diff
@@ -430,23 +412,27 @@ def deflector_magnitude(self, band):
...
+        """
+        lens_model_list, kwargs_lens =  self.deflector_mass_model_lenstronomy()
+        image_pos = self.point_source_image_positions()
+        arrival_times_list = []
+        for index, source in enumerate(self.source):
+            lens_model = LensModel(
+                lens_model_list=lens_model_list,
+                cosmo=self.cosmo,
+                z_lens=self.deflector_redshift,
+                z_source=source.redshift,
```
sibirrer, line 428: we have to feed it z_source_convention otherwise this does not work and the lens model parameters are interpreted to the specific source redshift (which it is not, in general)

slsim/lens.py
```diff
@@ -517,69 +523,92 @@ def extended_source_magnitude(self, band, lensed=False):
...
+                _light_model_list = kwargs_model.get(
+                        "source_light_model_list", [])[index]
+                kwargs_source_mag = [kwargs_params["kwargs_source"][index]]
+                if isinstance(_light_model_list, list):
+                    light_model_list = _light_model_list
+                else:
+                    light_model_list = [_light_model_list]
+                lightModel = LightModel(
+                    light_model_list=light_model_list)
+                lensModel = LensModel(
```
sibirrer, line 581: the LensModel() instance needs to know the z_source_convention and z_source in it

tests/test_lens.py
```diff
@@ -373,14 +430,85 @@ def test_image_number(self):
...
+
+def test_double_sersic_multisource(supernovae_lens_instance_double_sersic_multisource):
+    lens_class = supernovae_lens_instance_double_sersic_multisource
+    results = lens_class.source_light_model_lenstronomy(band="i")
+    assert len(results[0]["source_light_model_list"]) == 2
+    assert len(results[1]["kwargs_source"]) == 2
+    assert len(results[1]["kwargs_ps"]) == 2
 
 
+
```
sibirrer, line 512: I'd like to see the following test for the Lens() class:
1. Initiate two Lens() classes with each the same deflector and different point+extended sources at two different redshift
2. Initiate a third Lens() class with the same deflector and both sources at the same time
3. Then check that all the properties, such as time delays, magnifications, extended source magnifications etc are identical between the the separate and the combined classes.

sibirrer (changes requested):
Thank you very much!
We need to update the Lens() classes with the specific redshift to make it work with multiple source redshifts. I hopefully can push the changes tomorrow to make a new release of lenstronomy

nkhadka21:
Thank you very much @sibirrer ! Waiting for the lenstronomy update.

slsim/lens.py
```diff
@@ -341,25 +370,45 @@ def external_shear(self):
...
+        self._theta_E_list = []
+        for source in self.source:
+            self._theta_E_list.append(self._einstein_radius_deflector(source))
+        return self._theta_E_list
+
+    def _einstein_radius_deflector(self, source):
+        """Einstein radius, from SIS approximation (coming from velocity dispersion)
+        without line-of-sight correction.
+
+        :return: einstein radius of a lens-source pair.
```
sibirrer, line 390: docstring for source missing, in particular using the redshift of this source

slsim/lens.py
```diff
@@ -371,9 +420,21 @@ def einstein_radius(self):
         """Einstein radius, from SIS approximation (coming from velocity dispersion) +
         external convergence effect.
 
+        :return: list of Einstein radius [arc seconds] for each lens source pair.
+        """
+        theta_E_list = []
+        for source in self.source:
+            theta_E_list.append(self._einstein_radius(source=source))
```
sibirrer, line 427: do we want to save the calculations in cache, i.e. save it if available?

slsim/lens.py
```diff
@@ -593,11 +776,19 @@ def lenstronomy_kwargs(self, band=None):
...
+        # list of 
         kwargs_model = {
             "lens_light_model_list": lens_light_model_list,
             "lens_model_list": lens_mass_model_list,
         }
+        if self.source_number > 1:
+            kwargs_model["lens_redshift_list"] = [
+                self.deflector_redshift]*len(lens_mass_model_list)
+            kwargs_model["z_lens"] = self.deflector_redshift
+            kwargs_model["z_source"] = self.max_redshift_source_class.redshift
```
sibirrer, line 788: can we remove this part?

tests/test_roman_image_simulation.py
```diff
@@ -46,10 +47,19 @@
 }
 
 BAND = "F106"
-
+source = Source(
```
sibirrer, line 50: can we make a sharp image test by adding the two separate images resulting in the same as generating a single call for both sources (modulo the lens light)

slsim/lens.py
```diff
@@ -167,75 +126,144 @@ def deflector_position(self):
...
+            return validity_results
 
-    def validity_test(
+    def _validity_test(
         self,
+        source,
         min_image_separation=0,
         max_image_separation=10,
         mag_arc_limit=None,
+        source_index=None
```
sibirrer, line 264: can you document this input?

sibirrer (approved):
Looks great @nkhadka21 . I have one minor comment but already approve this PR

PR #273, opened 2024-11-07
False positive routine

nkhadka21:


slsim/FalsePositives/false_positive.py
```diff
@@ -0,0 +1,74 @@
+import numpy as np
+from slsim.lens import Lens
+
+class FalsePositive(Lens):
+    """Class to manage individual false positive."""
```
sibirrer, line 5: can you please describe a bit more what specifically 'false positives' are here

slsim/FalsePositives/false_positive_pop.py
```diff
@@ -0,0 +1,116 @@
...
+        self,
+        elliptical_galaxy_population: DeflectorsBase,
+        blue_galaxy_population: SourcePopBase,
+        cosmo: Optional[Cosmology] = None,
+        los_config: Optional[LOSConfig] = None,
+        source_number_choice: Optional[list[int]] = [1, 2, 3],
+        weights_for_source_number: Optional[list[int]] = None,
+    ):
+        """
+        Args:
```
sibirrer, line 26: formatting a bit odd. Also, please provide a better description of what the class does

slsim/FalsePositives/false_positive_pop.py
```diff
@@ -0,0 +1,116 @@
...
+    def draw_false_positive(self, number=1):
+        """Draw given number of false positives within the cuts of the lens and source.
+
+        :param number: number of false positive requested. The default value is 1.
+        :return: list of FalsePositive() instance with parameters of a pair of
+            elliptical and blue galaxy.
+        """
+        false_positive_population = []
+
+        for _ in range(number):
```
sibirrer, line 56: this is a rather long definition. Can you split it into two or more definitions?

slsim/Sources/galaxies.py
```diff
@@ -120,14 +120,24 @@ def source_number_selected(self):
...
 
+        :param z_max: maximum redshift limit for the galaxy to be drawn. If no galaxy is
+          found for this limit, None will be returned.
         :return: dictionary of source
         """
-
-        index = random.randint(0, self._num_select - 1)
-        galaxy = self._galaxy_select[index]
+        if z_max is not None:
+            filtered_galaxies = self._galaxy_select[self._galaxy_select["z"] < z_max]
```
sibirrer, line 132: for generating a large set of false positives, is this slowing the calculation down? You have to filter a potentially very long list first each time you draw a false positive. Just checking

sibirrer (changes requested):
Thank you @nkhadka21. I have some requests and comments about documentation and structure.

slsim/Sources/galaxies.py
nkhadka21: Yes, this slows down the calculation. z_max depands on each deflector that we draw. So, we need to do this here for each individual deflector. Currently, it takes 10 sec to draw 2000 false positive sample.

slsim/FalsePositives/false_positive_pop.py
```diff
@@ -0,0 +1,124 @@
...
+from slsim.ParamDistributions.los_config import LOSConfig
+from slsim.Deflectors.deflectors_base import DeflectorsBase
+from slsim.Sources.source import Source
+from slsim.Deflectors.deflector import Deflector
+from slsim.lens_pop import draw_test_area
+import random
+
+
+class FalsePositivePop(object):
+    """Class to perform samples of false positive population. Here, Here, false
```
sibirrer, line 14: typo in docstring: "Here, Here..."

slsim/FalsePositives/false_positive_pop.py
```diff
@@ -0,0 +1,124 @@
...
+        self.los_config = los_config
+        if self.los_config is None:
+            self.los_config = LOSConfig()
+    
+    def draw_deflector(self):
+        """Draw and prepare a deflector (lens) with tolerance-based z_max.
+        
+        :return: a deflector instance and deflector redshift with tolerance added.
+        """
+        lens = self._lens_galaxies.draw_deflector()
```
sibirrer, line 57: can't we have a definition in _lens_galaxies that returns the deflector object, instead of the Lens() object? Seems to be a weird way to initialize a Deflector() class from a Lens() class

slsim/FalsePositives/false_positive_pop.py
```diff
@@ -0,0 +1,124 @@
...
+        """Draw source(s) within the redshift limit of z_max.
+        :param z_max: maximum redshift for drawn source.
+        :return: A Source instance or a list of Source instance.
+
+        """
+        source_number = random.choices(self._choice, weights=self._weights)[0]
+        source_list = []
+
+        for _ in range(source_number):
+            source = self._sources.draw_source(z_max=z_max)
```
sibirrer, line 75: same as for deflectors, this is a bit a work-around to get to a Source() class. I would consider to make an instance of a Source() class, one does not need to go through this in this class

slsim/FalsePositives/false_positive_pop.py
```diff
@@ -0,0 +1,124 @@
...
+        :param number: number of false positive requested. The default value is 1.
+        :return: list of FalsePositive() instance.
+         """
+        false_positive_population = []
+
+        for _ in range(number):
+            successful = False
+            while not successful:
+                # Step 1: Draw deflector
+                lens, z_max = self.draw_deflector()
```
sibirrer, line 106: I would call the object 'deflector' and not 'lens' here

slsim/FalsePositives/false_positive_pop.py
```diff
@@ -0,0 +1,124 @@
...
+                # Step 2: Draw sources
+                source = self.draw_sources(z_max)
+                if source is None:
+                    continue  # Retry if sources are invalid
+
+                # Step 3: Create false positive
+                vd=lens.velocity_dispersion(cosmo=self.cosmo)
+                test_area = 3 * draw_test_area(
```
sibirrer, line 114: is this a bit arbitrary? The area and the interloper space density might be chosen by the user instead.

sibirrer (changes requested):
Thank you! Now, that's a bit more clear to me, I have some other suggestions to make the code a bit more streamlined.

sibirrer (approved):
Looks great, thank you very much!

PR #276, opened 2024-11-14
Einstein radius query simplified

sibirrer:
There was a distinction between Einstein radius of the deflector-only and including the external convergence. This was done not fully correctly as the numerical calculation did not correct for the external convergence, and the EPL model has a power-law dependent correction of the Einstein radius with external convergence.
To make things simpler, I removed the deflector Einstein radius from the API and changed the EPL Einstein radius calculation correcting for the kappa_ext.


nkhadka21 (approved):
Hi @sibirrer , thank you very much for this PR! It looks good to me.

PR #277, opened 2024-11-15
refactoring LOS classes and split into individual LOS and population

sibirrer:
I split the LOS classes and demand the Lens() object to input a LOSIndividual() class

### DIFF
--- a/slsim/lensed_system_base.py
+++ b/slsim/lensed_system_base.py
@@ -1,22 +1,26 @@
 from abc import ABC, abstractmethod
 
+
 class LensedSystemBase(ABC):
     """Abstract Base class to create a lens system with all lensing properties required
     to render populations."""
 
-    def __init__(self, source_class, deflector_class):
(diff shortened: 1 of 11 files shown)


slsim/lensed_system_base.py
```diff
@@ -1,22 +1,26 @@
 from abc import ABC, abstractmethod
 
+
 class LensedSystemBase(ABC):
     """Abstract Base class to create a lens system with all lensing properties required
     to render populations."""
 
-    def __init__(self, source_class, deflector_class):
+    def __init__(self, source_class, deflector_class, los_class):
```
nkhadka21, line 8: Input los_class is not used. Do you think we should define a variable using this input?

nkhadka21 (approved):
Hi @sibirrer , thank you very much for this PR! It looks good to me, and I approve it. However, I have a minor comment/query. Please look into it.

PR #280, opened 2024-11-18
Added functionality to convert MJD time to days for transient events.

nkhadka21:


sibirrer (approved):
Thank you @nkhadka21! I have a request for re-naming what you called zero_point. Otherwise is looking good. Would be great if you could also update/add a notebook with an OpSimSummary created light curve with a pre-defined event time

nkhadka21:
Hi @sibirrer, I changed the name you suggested and also added a notebook showing use of opsim_pipeline.

PR #285, opened 2024-11-21
More robust for LOS

trivialTZ:
if z_source <= z_lens, LOS can still be passed.


sibirrer (approved):
Thank yo @trivialTZ !

PR #287, opened 2024-11-22
Fixed multi-source plane conventions and solver and added EPL velocity dispersions

sibirrer:
Fixed a bug with the multi-source plane analytical lens equation solver, added routines to self-consistently related velocity dispersion and power-law slopes.

Also debugged a recently introduced bug on the line of sight function output interpretation


nkhadka21 (approved):
Hi @sibirrer, thank you very much for this PR! It looks good to me.

PR #288, opened 2024-11-22
Fixed some bug in multi source lensing and los.

nkhadka21:


slsim/lens.py
```diff
@@ -376,7 +376,7 @@ def los_linear_distortions(self):
         """
         kappa = self.los_class.convergence
         gamma1, gamma2 = self.los_class.shear
-        return kappa, gamma1, gamma2
+        return gamma1, gamma2, kappa
```
sibirrer, line 379: I'd like to have it defined the opposite way with kappa first (see also the docstrings in this routine

slsim/lens.py
```diff
@@ -355,7 +355,7 @@ def deflector_redshift(self):
 
         :return: lens redshift
         """
-        return self.deflector.redshift
+        return float(self.deflector.redshift)
```
sibirrer, line 358: I already made changes in my PR for this. I'd like to have the self.deflector.redshift quantity well defined as float, not the return of the function in lens.py (same for the source redshifts as well

slsim/lens.py
```diff
@@ -732,6 +732,12 @@ def _extended_single_source_magnification(self, source, source_index):
                 source_index
             ]
             kwargs_source_mag = kwargs_params["kwargs_source"][source_index]
+        if isinstance(light_model_list, list):
```
sibirrer, line 735: I don't think we need those things as the components should already be defined as lists

sibirrer (changes requested):
Thank you @nkhadka21. I hope you saw my PR that addresses most of these things. I have the tests working and have also some more changes implemented (including the minor bug in the lens equation solver etc)

slsim/lens.py
nkhadka21: Yes, I saw your PR! and changing things occordingly.

slsim/lens.py
nkhadka21: These lines are applicable while accessing keywords in some case (just one case where we have multisource and single sersic light profile. In this case above line will produce str and dictionary due to the indexing. So, need to convert them to list.)

PR #295, opened 2025-01-28
rescaled the galaxy size.

nkhadka21:
This PR accounts for the redshift evolution in galaxy size.
f04ff4c9-c1f4-4dfb-a502-f0e848e46c73

### DIFF
--- a/slsim/Util/param_util.py
+++ b/slsim/Util/param_util.py
@@ -451,8 +451,9 @@ def vel_disp_from_m_star(m_star):
 
 
 def galaxy_size_redshift_evolution(z):
-    """This function provides a galaxy size elolution with the redshift. This 
-     relation is taken from Shibuya et al. (2015): https://arxiv.org/abs/1503.07481
+    """This function provides a galaxy size elolution with the redshift.
+    Provides a correction factor to the size relative to z=0.
+    This relation is taken from Shibuya et al. (2015): https://arxiv.org/abs/1503.07481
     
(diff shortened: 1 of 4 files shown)


slsim/Util/param_util.py
```diff
@@ -448,3 +448,16 @@ def vel_disp_from_m_star(m_star):
     """
     v_disp = np.power(10, 2.32) * np.power(m_star / 1e11, 0.24)
     return v_disp
+
+
+def galaxy_size_redshift_evolution(z):
+    """This function provides a galaxy size which evolves with the redshift. This
+     relation is taken from https://iopscience.iop.org/article/10.1088/0067-0049/219/2/15/pdf
```
sibirrer, line 455: I would put the Author et al in the description as well and link to the arXiv landing page, not the pdf.
Also, the description of the routine is not accurate. This is specifically a correction factor on the size relative to z=0

slsim/Util/param_util.py
```diff
@@ -448,3 +448,16 @@ def vel_disp_from_m_star(m_star):
...
+
+    :param z: galaxy redshift. It can be a single galaxy redshift or list of galaxy
+     redshifts.
+    :return: Physical size of the galaxy.
+    """
+    Bz = 6.9
```
sibirrer, line 461: can you also document specifically from where these numbers are (i.e. Table X in paper etc)

sibirrer (changes requested):
Thank you very much! I have some minor documentation requests

PR #296, opened 2025-01-30
bug fix in ellipticity scaling

sibirrer:
the mass to light ellipticity scaling should operate on the same definition of either. Previously it was not done. This change, which makes it consistent, will, however, also change the ellipticity distributions of galaxies, which should be looked into


nkhadka21:
Hi @sibirrer , thank you very much for this fix! It looks good to me but CI is failing.

nkhadka21 (approved):

PR #297, opened 2025-02-03
Lensed source mag cut

nkhadka21:
This PR implements an option to apply magnitude on second brightest image.

### DIFF
--- a/slsim/lens.py
+++ b/slsim/lens.py
@@ -267,6 +267,7 @@ class Lens(LensedSystemBase):
         min_image_separation=0,
         max_image_separation=10,
         mag_arc_limit=None,
+        second_bright_image_cut=None,
         source_index=None,
     ):
         """Check whether a single lensing configuration matches selection and
@@ -278,6 +279,10 @@ class Lens(LensedSystemBase):
             magnitude limits of integrated lensed arc
         :type mag_arc_limit: dict with key of bands and values of
             magnitude limits
+        :param second_bright_image_cut: Dictionary containing maximum magnitude of the 
+         second brightest image and corresponding band. If provided, selects lenses 
+         where the second brightest image has a magnitude less than or equal to 
+         provided magnitude.
         :param source_index: index of a source in source list.
         :return: boolean
         """
@@ -349,6 +354,19 @@ class Lens(LensedSystemBase):
             if bool_mag_limit is False:
                 return False
         # TODO make similar criteria for point source magnitudes
+        # Criteria 7: (optional)
+        # computes the magnitude of each image and if the second brightest image has 
+        # the magnitude less or equal to second_bright_mag_max.
+        if second_bright_image_cut is not None:
+            if self._source_type == "extended":
+                image_magnitude_list = self.extended_source_magnitude_for_each_images(
+                    band=second_bright_image_cut["band"], lensed=True)
+            elif self._source_type in ["point_plus_extended", "point_source"]:
+                image_magnitude_list = self.point_source_magnitude(
+                    band=second_bright_image_cut["band"], lensed=True)
+            second_bright_mag = np.sort(image_magnitude_list[0])[1]
+            if second_bright_mag > second_bright_image_cut["second_bright_mag_max"]:
+                return False
         return True
         # TODO: test for signal-to-noise ratio in surface brightness
 
@@ -629,6 +647,25 @@ class Lens(LensedSystemBase):
                     magnified_mag_list.append(source_mag_unlensed - magnif_log[i])
                 return np.array(magnified_mag_list)
         return source.point_source_magnitude(band)
+    
+    def extended_source_magnitude_for_each_images(self, band, lensed=False):
+        """extended source magnitudes, either unlensed (single value) or lensed
+        (array) with macro-model magnifications. This function provided
+        magnitudes of all the sources.
+
+        :param band: imaging band
+        :type band: string
+        :param lensed: if True, returns the lensed magnified magnitude of each images.
+        :type lensed: bool
+        :return: list of extended source magnitudes.
+        """
+
+        magnitude_list = []
+        for source in self.source:
+            magnitude_list.append(
+                self._extended_source_magnitude_for_each_images(band, source,
+                                                                 lensed=lensed))
+        return magnitude_list
 
     def extended_source_magnitude(self, band, lensed=False):
         """Unlensed apparent magnitude of the extended source for a given band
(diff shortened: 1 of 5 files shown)


slsim/lens.py
```diff
@@ -234,7 +234,11 @@ def _point_source_image_positions(self, source):
         return self._point_image_positions
 
     def validity_test(
-        self, min_image_separation=0, max_image_separation=10, mag_arc_limit=None
+        self,
+        min_image_separation=0,
+        max_image_separation=10,
+        mag_arc_limit=None,
+        second_bright_image_cut=None,
```
sibirrer, line 241: I would rename it to second_brightest_image_cut

slsim/lens.py
```diff
@@ -245,6 +249,12 @@ def validity_test(
             magnitude limits of integrated lensed arc
         :type mag_arc_limit: dict with key of bands and values of
             magnitude limits
+        :param second_bright_image_cut: Dictionary containing maximum
+            magnitude of the second brightest image and corresponding
+            band. If provided, selects lenses where the second brightest
+            image has a magnitude less than or equal to provided
+            magnitude. eg: second_bright_image_cut = {"band": "i",
+            "second_bright_mag_max": 23}
```
sibirrer, line 257: I don't think you need to name the parameter itself "second_bright_mag_max" since the entire dictionary is already labeled this way. Perhaps simply "mag_max" or something like that.

slsim/lens.py
```diff
@@ -349,6 +367,22 @@ def _validity_test(
             if bool_mag_limit is False:
                 return False
         # TODO make similar criteria for point source magnitudes
+        # Criteria 7: (optional)
+        # computes the magnitude of each image and if the second brightest image has
+        # the magnitude less or equal to "second_bright_mag_max" provided in the dict
+        # second_bright_image_cut.
+        if second_bright_image_cut is not None:
+            if self._source_type == "extended":
+                image_magnitude_list = self.extended_source_magnitude_for_each_images(
```
sibirrer, line 376: typo in english: extended_source_magnitude_for_each_image() without 's'

slsim/lens.py
```diff
@@ -349,6 +367,22 @@ def _validity_test(
...
+                    band=second_bright_image_cut["band"], lensed=True
```
sibirrer, line 377: If I understand correctly, this function only applies a cut to one band only?

slsim/lens.py
```diff
@@ -349,6 +367,22 @@ def _validity_test(
...
+                )
+            elif self._source_type in ["point_plus_extended", "point_source"]:
+                image_magnitude_list = self.point_source_magnitude(
+                    band=second_bright_image_cut["band"], lensed=True
+                )
+            second_bright_mag = np.sort(image_magnitude_list[0])[1]
```
sibirrer, line 383: is sorting ascending or descending? Did you do a test checking this is the second most bright image?

slsim/lens.py
```diff
@@ -630,6 +664,28 @@ def _point_source_magnitude(self, band, source, lensed=False, time=None):
                 return np.array(magnified_mag_list)
         return source.point_source_magnitude(band)
 
+    def extended_source_magnitude_for_each_images(self, band, lensed=False):
```
sibirrer, line 667: typo in naming

slsim/lens.py
```diff
@@ -630,6 +664,28 @@ def _point_source_magnitude(self, band, source, lensed=False, time=None):
...
+        """Extended source magnitudes, either unlensed (single value) or lensed
+        (array) with macro-model magnifications. This function provided
```
sibirrer, line 669: this is an inaccurate description. You need to state that this is when assuming all the light is concentrated at the center of the extended image and magnified as a point source multiple times

slsim/lens.py
```diff
@@ -652,10 +708,33 @@ def extended_source_magnitude(self, band, lensed=False):
...
+
+        :param band: imaging band
+        :type band: string
+        :param lensed: if True, returns the lensed magnified magnitude
+            of each image.
+        :type lensed: bool
+        :return: extended source magnitude of a single source.
+        """
+        if lensed:
+            magnif = self._point_source_magnification(source)
```
sibirrer, line 724: will this solve the lens equation? The reason I ask is whether there is already a solution for the lens equation available from other routines. I just like to avoid multiple times solving the lens equation for the same source

slsim/lens.py
```diff
@@ -688,14 +767,16 @@ def point_source_magnification(self):
                 )
         return self._ps_magnification_list
 
-    def _point_source_magnification(self, source):
+    def _point_source_magnification(self, source, extended=False):
         """Macro-model magnification of a point source. This is for a single
-        source.
+        source. The function also works for extended source.
```
sibirrer, line 772: Be more explicit: state that it will use the center of the extended source instead to calculate the lensing magnifications

slsim/lens_pop.py
```diff
@@ -57,6 +59,12 @@ def select_lens_at_random(self, test_area=None, **kwargs_lens_cut):
         :param test_area: solid angle around one lensing galaxies to be
             investigated on (in arc-seconds^2). If None, computed using
             deflector's velocity dispersion.
+        :param second_bright_image_cut: Dictionary containing maximum
```
sibirrer, line 62: can't this parameter just be as one of the kwargs_lens_cut?

slsim/lens_pop.py
```diff
@@ -122,7 +132,13 @@ def get_num_sources_tested(self, testarea=None, num_sources_tested_mean=None):
         num_sources_range = np.random.poisson(lam=num_sources_tested_mean)
         return num_sources_range
 
-    def draw_population(self, kwargs_lens_cuts, multi_source=False, speed_factor=1):
+    def draw_population(
+        self,
+        kwargs_lens_cuts,
+        multi_source=False,
+        second_bright_image_cut=None,
```
sibirrer, line 139: dito, can't this parameter be absorbed in kwargs_lens_cut?

sibirrer (changes requested):
Thank you @nkhadka21 ! I have some requests for changes and questions about potentially improving the design.

slsim/lens.py
nkhadka21: Yes, you are correct.

slsim/lens.py
nkhadka21: sorting is ascending. I had checked this.

slsim/lens.py
nkhadka21: Solution to the lens equation is obtained from other routine. Lens equation is not directly solved within the _point_source_magnification(source) function.

sibirrer (approved):
Thank you @nkhadka21 !

PR #298, opened 2025-02-06
minor notebook fix and increase redshift grid precision.

nkhadka21:
This PR has fixed following:
1. minor update in supernovae overview notebook.
2. Increases the redshift grid precision in supernovae lightcone integration.
3. Reduces the redshift width in supernovae host match.


sibirrer (approved):
Thank you @nkhadka21 !

PR #301, opened 2025-02-13
correction to the poisson noise.

nkhadka21:
This PR improves the Poisson noise by accounting for the difference between the magnitude zero point of the coadd and single visit, and also accounts for the CCD gain.

### DIFF
--- a/slsim/image_simulation.py
+++ b/slsim/image_simulation.py
@@ -503,7 +503,9 @@ def image_plus_poisson_noise(
         27).
     :param single_visit_zero_point: Zero point of the single-visit image
         (default 27 for g-band).
-    :return: image with possion noise
+    :return: image with possion noise. The function returns ADU/sec in all cases, 
+     regardless of whether gain = 1 or not. The noise is applied in the electron domain,
+     but the final image is converted back to ADU/sec.
     """
     # make sure all values in an image are positive
     image_positive = np.clip(image, 0, None)
--- a/slsim/image_simulation.py
+++ b/slsim/image_simulation.py
@@ -490,17 +490,19 @@ def deflector_images_with_different_zeropoint(
     return image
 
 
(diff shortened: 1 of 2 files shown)


sibirrer:
tests are failing @nkhadka21

nkhadka21:
> tests are failing @nkhadka21

This is strange! I have not touched the code which is failing. It passes while I test it locally. So, can not even reproduce this failure. Trying to fix it.

nkhadka21:
@sibirrer , this PR is ready for the review!

slsim/image_simulation.py
```diff
@@ -490,16 +490,32 @@ def deflector_images_with_different_zeropoint(
...
+    # make sure all values in an image are positive
+    image_positive = np.clip(image, 0, None)
+    # Compute zero point scaling factor
+    zero_point_scale = 10 ** (0.4 * (coadd_zero_point - single_visit_zero_point))
+    # Convert counts-per-second to electrons
+    cps_to_electrons = exposure_time * gain / zero_point_scale
+    # get electron count with poisson noise
+    noisy_electrons = np.random.poisson(image_positive * cps_to_electrons)
+    # convert back to count per sec
+    noisy_image = noisy_electrons / cps_to_electrons
```
sibirrer, line 517: in case the Gain is !=1, do you expect to return electrons/second or ADU/second? Basically, is the units of the noise with Gain or without? This perhaps should be specified in the return statement

sibirrer (commented):
Looks good, I have one relatively minor suggestion

slsim/image_simulation.py
nkhadka21: The function returns ADU/sec in all cases, regardless of whether gain = 1 or not. The noise is applied in the electron domain, but the final image is converted back to ADU/sec.

sibirrer (approved):
Thank you, makes sense

PR #303, opened 2025-03-03
update gg-lensing tutorial

nkhadka21:


sibirrer (approved):
Thank you @nkhadka21 !

PR #305, opened 2025-03-05
update on lsst science pipeline module

nkhadka21:
This PR implements,
1. an option to provide list of lenses in lens injection
2. coadd degrade from 5 year to any years which will be used to degrade 5 year dp0 coadd images to 1 year coadd.


sibirrer (approved):
Looks great - thank you!

PR #306, opened 2025-03-10
Source size correction.

nkhadka21:
This PR implements a source size from Bernardi et al 2003. Also, added a function to generate lens id.

### DIFF
--- a/slsim/Sources/galaxies.py
+++ b/slsim/Sources/galaxies.py
@@ -303,16 +303,29 @@ def convert_to_slsim_convention(
     size in skypy source catalog to arcsec.
 
     :param galaxy_catalog: An astropy table of galaxy catalog in other
-     conventions. 
-    :type galaxy_catalog: astropy Table object. 
+    conventions. :type galaxy_catalog: astropy Table object. :param
+    light_profile: keyword for number of sersic profile to use in
+    source light model. accepted kewords: "single_sersic",
+    "double_sersic". :param input_catalog_type: type of the catalog. If
+    someone wants to use scotch  catalog or skypy catalog, they need to
+    specify it. :type input_catalog_type: str. eg: "scotch" or "skypy".
+    :param source_size: If "Bernardi", computes galaxy size  using
+    g-band magnitude otherwise rescales skypy source size to Shibuya et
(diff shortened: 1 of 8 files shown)


slsim/Sources/galaxies.py
```diff
@@ -292,12 +302,31 @@ def convert_to_slsim_convention(
     convert scotch catalog to slsim convension and to change unit of angular
     size in skypy source catalog to arcsec.
 
+    :param galaxy_catalog: An astropy table of galaxy catalog in other
```
sibirrer, line 305: The formatting has been messed up here

tests/test_param_util.py
```diff
@@ -383,5 +385,20 @@ def test_degrade_coadd_data():
...
+    mapp = 24.0  # Apparent g-band magnitude
+    zsrc = 0.5  # Source redshift
+
+    # Call function
+    Reff, Reff_arcsec = galaxy_size(mapp, zsrc, cosmo)
+
+    # Check outputs are finite and positive
+    assert np.isfinite(Reff) and Reff > 0
+    assert np.isfinite(Reff_arcsec) and Reff_arcsec > 0
+    assert Reff == 0.5322278567954598
```
sibirrer, line 399: perhaps we can test this with a specified numerical precision?

sibirrer (changes requested):
Thank you very much! I have two minor comments

slsim/Sources/galaxies.py
sibirrer: still messed up format

sibirrer (changes requested):
the doctoring formatting still does not work

nkhadka21:
> the doctoring formatting still does not work

@sibirrer , I think it is fixed now!

slsim/Sources/galaxies.py
sibirrer: no, it's still not fixed

sibirrer (changes requested):

nkhadka21:
@sibirrer , I think it is fixed now!

sibirrer (approved):

PR #308, opened 2025-03-11
Central source check

nkhadka21:
This PR implements a method to check if an image contains a bright source within the central box of a specified dimension. This is used in the LSST Science Pipeline source injection task. It ensures that our injected lens does not overlap with an existing object.

### DIFF
--- a/slsim/LsstSciencePipeline/lsst_science_pipeline.py
+++ b/slsim/LsstSciencePipeline/lsst_science_pipeline.py
@@ -12,7 +12,6 @@ from slsim.image_simulation import point_source_coordinate_properties
 from slsim.Util.param_util import (
     random_ra_dec,
     fits_append_table,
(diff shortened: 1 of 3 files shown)


sibirrer (commented):
Overall looks good. I am a bit worried about the size of the definition that extends further. Perhaps you can write sub-definitions to make it more structured? Or at least add some more comments in the code? I am worried for future developments that it is hard to change and read the code

nkhadka21:
> Overall looks good. I am a bit worried about the size of the definition that extends further. Perhaps you can write sub-definitions to make it more structured? Or at least add some more comments in the code? I am worried for future developments that it is hard to change and read the code

Hi @sibirrer , I refactored this function! It is now split into three separate functions.

sibirrer (approved):
Thank you very much! Looks good (pending tests)

PR #309, opened 2025-03-17
connected FalsePositivePop with the lsst_science_pipeline module.

nkhadka21:
This PR connects FalsePositivePop with the lsst_science_pipeline module. Additionally, it changes the convention of y band magnitude from mag_Y to mag_y to be consistent with the DP0 data.

### DIFF
--- a/slsim/LsstSciencePipeline/lsst_science_pipeline.py
+++ b/slsim/LsstSciencePipeline/lsst_science_pipeline.py
@@ -300,7 +300,9 @@ def lens_inejection_fast(
         detection in center box (default is 5).
     :param false_positive: Boolean. If false, code assumes that the provided
      population is a lens population. If True, code assumes that the provided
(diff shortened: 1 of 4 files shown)


slsim/LsstSciencePipeline/lsst_science_pipeline.py
```diff
@@ -297,6 +298,9 @@ def lens_inejection_fast(
         is 3 arcsec).
     :param center_source_snr_threshold: SNR threshold for object
         detection in center box (default is 5).
+    :param false_positive: Boolean. If false, code assumes that the provided
+     population is a lens population. If True, code assumes that the provided
```
sibirrer, line 302: this doctoring might need some more explanation or pointer to a specific function of what it does

sibirrer (approved):
Looks good. I have one minor comment.

PR #311, opened 2025-03-19
added contrast ratio of lens and source surface brightness,

sibirrer:
changed definition Lens.source() to access individual sources with input instead of being a list


nkhadka21 (approved):
Hi @sibirrer , thank you very much for this PR! It looks good to me.

PR #312, opened 2025-03-21
Source distribution update

nkhadka21:
Changed the polynomial coefficient of the M* vs z relation to shift the redshift distribution peak between redshifts 1 and 2. Also, adjusted the normalization coefficient to control the total number of galaxies.


sibirrer (approved):
I guess you have compared these distributions to still match the low-redshift normalizations?

PR #314, opened 2025-03-28
Source refactor

nkhadka21:
This PR refactors Source class!


slsim/Sources/SourceTypes/extended_source.py
```diff
@@ -0,0 +1,160 @@
...
+            e1 = \\frac{1-q}{1+q} * cos(2 \\phi)
+            e2 = \\frac{1-q}{1+q} * sin(2 \\phi)
+
+        with q being the minor-to-major axis ratio.
+        """
+
+        return self._source.ellipticity
+
+    @property
+    def n_sersic(self):
```
sibirrer, line 65: is this a property that you need access to at this level?

slsim/Sources/SourceTypes/extended_source.py
```diff
@@ -0,0 +1,160 @@
...
+        """Returns sersic indices of the source profile."""
+
+        return self._source.n_sersic
+
+    @property
+    def sersicweight(self):
```
sibirrer, line 71: is this a property that you need access to at this level?

slsim/Sources/SourceTypes/extended_source.py
```diff
@@ -0,0 +1,160 @@
...
+        """Returns weight of the sersic components."""
+
+        return self._source.sersicweight
+
+    @property
+    def image_redshift(self):
```
sibirrer, line 77: I don't think you need access to this very specific thing of one particular source type. Find a way you can retrieve the information you need from the source without creating this access level

slsim/Sources/SourceTypes/extended_source.py
```diff
@@ -0,0 +1,160 @@
...
+        """Returns redshift of a given image."""
+
+        return self._source.image_redshift
+
+    @property
+    def image(self):
```
sibirrer, line 83: dito

slsim/Sources/SourceTypes/extended_source.py
```diff
@@ -0,0 +1,160 @@
...
+        """Returns image of a given extended source."""
+
+        return self._source.image
+
+    @property
+    def phi(self):
```
sibirrer, line 89: what 'image' is meant here? This is also too specific that you might not need access to

slsim/Sources/SourceTypes/extended_source.py
```diff
@@ -0,0 +1,160 @@
...
+        """Returns position angle of a given image in arcsec."""
+
+        return self._source.phi
+
+    @property
+    def pixel_scale(self):
```
sibirrer, line 95: dito

slsim/Sources/SourceTypes/extended_source.py
```diff
@@ -0,0 +1,160 @@
...
+
+        :param reference_position: reference position. the source postion will be
+         defined relative to this position.
+         Eg: np.array([0, 0])
+        :param draw_area: The area of the test region from which we randomly draw a source
+            position. Eg: 4*pi.
+        :return: [x_pos, y_pos]
+        """
+
+        return self._source.extended_source_position(reference_postion, draw_area)
```
sibirrer, line 115: why drawing the source position in all individual source classes? This would be the class that manages these choices that are independent of the individual source classes. I.e. make the draw in this class and then send it to lower level, if needed

slsim/Sources/SourceTypes/extended_source.py
```diff
@@ -0,0 +1,160 @@
...
+
+        return self._source.phi
+
+    @property
+    def pixel_scale(self):
+        """Returns pixel scale of a given image."""
+
+        return self._source.pixel_scale
+
+    def extended_source_position(self, reference_postion, draw_area):
```
sibirrer, line 100: this routine should also be able to be executed without specifying the inputs in case the source position already exists. Perhaps make the arguments =None as defaults

slsim/Sources/SourceTypes/interpolated_image.py
```diff
@@ -0,0 +1,108 @@
+from slsim.Sources.SourceTypes.source_base import SourceBase
+from slsim.Util.cosmo_util import z_scale_factor
+
+
+class Interpolated(SourceBase):
```
sibirrer, line 5: think about which definition need to be public and make the other ones private

slsim/Sources/SourceTypes/point_plus_extended_source.py
```diff
@@ -0,0 +1,32 @@
+from slsim.Sources.SourceTypes.point_source import PointSource
+from slsim.Sources.SourceTypes.extended_source import ExtendedSource
+
+
+class PointPlusExtendedSource(PointSource, ExtendedSource):
+    """Class to manage a single point source and a single extended source
+    (host)"""
+
+    def __init__(self, source_dict, cosmo=None, **kwargs):
```
sibirrer, line 9: for this class, I suggest having two dictionaries as inputs and two source types as inputs. It is unclear as a user how to initialize this class and what the conditions are for the two different inherited classes

slsim/Sources/SourceTypes/point_source.py
```diff
@@ -0,0 +1,105 @@
...
+from slsim.Sources.SourceTypes.quasar import Quasar
+from slsim.Sources.SourceTypes.general_lightcurve import GeneralLightCurve
+
+_SUPPORTED_POINT_SOURCES = ["supernova", "quasar", "general_lightcurve"]
+
+
+class PointSource(object):
+    """Class to manage a single point source."""
+
+    def __init__(self, source_dict, cosmo=None, **kwargs):
```
sibirrer, line 11: make point_source_type and perhaps all the parts that are explicitly being managed by this class explicit inputs to this class

slsim/lens.py
```diff
@@ -915,12 +925,14 @@ def lenstronomy_kwargs(self, band=None):
                 lens_mass_model_list
             )
             kwargs_model["z_lens"] = self.deflector_redshift
-            if self.max_redshift_source_class.light_profile in [
+            if self.max_redshift_source_class.kwargs["extendedsource_type"] in [
```
sibirrer, line 928: this is a bit an odd way to access this

slsim/lens.py
```diff
@@ -1135,10 +1140,17 @@ def generate_id(self, ra=None, dec=None):
             self._source_type == "point_source"
             or self._source_type == "point_plus_extended"
         ):
-            if self.max_redshift_source_class.sn_type is not None:
-                lens_type = "SN" + self.max_redshift_source_class.sn_type
-            else:
+            if self.max_redshift_source_class.kwargs["pointsource_type"] in [
```
sibirrer, line 1143: there might be easier way to access those quantities

sibirrer (changes requested):
Thank you very much @nkhadka21 ! This looks much better. I have quite a few comments that hopefully make the structure more explicit and structured.

slsim/Sources/SourceTypes/double_sersic.py
```diff
@@ -0,0 +1,170 @@
...
+        )
+
+    @property
+    def sersicweight(self):
+        """Returns weight of the sersic components."""
+
+        return self.source_dict["w0"], self.source_dict["w1"]
+
+    @property
+    def angular_size(self):
```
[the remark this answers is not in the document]
nkhadka21: Initially, the code was prepared for the scotch host galaxy catalog which has major-axis half-light size (arcsec) and minor-axis half-light size (arcsec) for Sersic component n=1 and n=2 separately. Then, we have computed these average angular size for each sersic component using their corresponding major and minor axis. This defination is only used internally within the class.

slsim/Sources/SourceTypes/extended_source.py
nkhadka21: Source position is only drawn in SourceBase class and is accessed in all individual classes. If someone wants to use these individual classes for their own purpose, dependency of these classes for source position on ExtendedSource class might create some difficuly, right?

slsim/Sources/SourceTypes/point_plus_extended_source.py
nkhadka21: Hi @sibirrer , I made pointsource_type and extendedsource_type as explicit inputs and also followed the same convention for ExtendedSource, PointSource,  and Source classes. However, for the source dict, we have followed passing single dict for a source at the top level. If we make change here, it will require splitting given source dict in two source dict which is extra step. Also, other part of the code is connected with it (eg: supernova catalog generation).

slsim/Sources/SourceTypes/double_sersic.py
```diff
@@ -0,0 +1,170 @@
...
+from slsim.Util.param_util import ellipticity_slsim_to_lenstronomy
+from slsim.Util.param_util import surface_brightness_reff
+
+
+class DoubleSersic(SourceBase):
+    """Class to manage source with double sersic light profile."""
+
+    def __init__(self, source_dict):
+        """
+        :param source_dict: Source properties. May be a dictionary or an Astropy table.
```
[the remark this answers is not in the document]
nkhadka21: Currently, both deflector and source types are following this way to access object properties. Maybe I can change both deflector and source type inputs in later time.

sibirrer (approved):
Looks good, thank you very much!
Can you please open issues for the features that we might change in the future?

nkhadka21:
> Looks good, thank you very much! Can you please open issues for the features that we might change in the future?

Thanks Simon! Yes, I will open issues.

PR #319, opened 2025-04-15
minor bug fix when having more than one deflector profile in estimation of Einstein radius

sibirrer:


nkhadka21 (approved):
Thanks @sibirrer for this quick fix! It looks good to me.

PR #328, opened 2025-05-17
Matching functionality from SkyPy sources to COSMOS real galaxy images

rahulmkarthik:
Added a matching function under Util that creates a table to map sersic galaxies to their nearest neighbors in real size and magnitude space.

Added a notebook to test the matching function and visualize the results.

Added a catalog merging function to astrometrically merge two catalogs with RA, DEC, and MAG columns.

### DIFF
--- a/slsim/Util/cosmo_util.py
+++ b/slsim/Util/cosmo_util.py
@@ -17,6 +17,34 @@ def merge_catalogs(
     sc_ra_col="RA",
     sc_dec_col="DEC",
 ):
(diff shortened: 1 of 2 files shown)


sibirrer (changes requested):
Thank you @rahulmkarthik ! Can you please add test functions and document strings of the definitions, input variables and outputs?

sibirrer (changes requested):
Thank you! Find my first batch of comments. Also, make either sure you use your definitions in the notebook (to make the notebook simpler) or you might not need some of the functions in SLSim

PR #329, opened 2025-05-19
simplify lenstronomy LensModel() class initialization

sibirrer:


sibirrer:
@nkhadka21 can we merge this one? There are other PRs that are coming in and I would like to have this (small) update being done before - thank you!

nkhadka21 (approved):

PR #331, opened 2025-05-22
Add Euclid congig option to slsim

FerroYx:
Add Euclid config from Lenstronomy to slsim for single lens simulations.


slsim/Observations/image_quality_lenstronomy.py
```diff
@@ -19,4 +20,6 @@ def kwargs_single_band(band, observatory="LSST", **kwargs):
         observatory = LSST(band=band, **kwargs)
     elif observatory == "Roman":
         observatory = Roman(band=band, **kwargs)
+    elif observatory == "Euclid":
```
sibirrer, line 23: can you add a test function for this? Thank you!

sibirrer:
@FerroYx can you check the pre-commit error messages - thank you!

sibirrer (approved):
Looks great, thank you very much @FerroYx 
@nkhadka21 might be worth thinking of adding the filters more manually to config files as we are duplicating currently a few yaml files just for the sake of changing the filters

FerroYx:
> Looks great, thank you very much @FerroYx @nkhadka21 might be worth thinking of adding the filters more manually to config files as we are duplicating currently a few yaml files just for the sake of changing the filters

Hi @nkhadka21 Please let me know if there is anything need to be updated!

nkhadka21 (commented):
Hi @FerroYx , thank you very much for this PR! It looks very good. I think you have used only 1 schechter function for red galaxies. It would be great if you could use two schechter function for red galaxies. Please see red galaxy part in following file: https://github.com/LSST-strong-lensing/slsim/blob/1d3609e30f6f047ef22cd2898a0845b550b19313/data/SkyPy/lsst-like_triple_SF.yml

nkhadka21:
> Looks great, thank you very much @FerroYx @nkhadka21 might be worth thinking of adding the filters more manually to config files as we are duplicating currently a few yaml files just for the sake of changing the filters

I agree with you. I think current slsim version accepts manual filter suply but I have not tested it very well. I will look into it. Thank you.

PR #332, opened 2025-05-27
Fixed a small typo in deflector.py

timedilatesme:
In the DeflectorTypes Base class, the units are in arcsec, so these should be maintained!


nkhadka21 (approved):
Thank you very much @timedilatesme! It looks good to me.

PR #337, opened 2025-06-17
change webbpsf to stpsf

ahuang314:
webbpsf was migrated to stpsf a few months ago. All of the code is backwards compatible, but we update slsim to use stpsf anyways.

**OLD:**
When using webbpsf, the user would have to manually download additional data files and change their operating system's environment variables to point towards the directory containing these webbpsf data files.

**NEW:**
stpsf also requires these additional data files, but the downloading is now done automatically upon calling stpsf for the first time. These data files are downloaded to /home/user/data, and the user no longer has to manually download them or change their environment variables.

TLDR less user overhead


ahuang314:
@sibirrer @nkhadka21
stpsf requires python 3.10 but slsim uses python 3.9. Do you want to update to using python 3.10 or stay on webbpsf?

sibirrer:
I think we can update to python 3.10 or higher

nkhadka21 (approved):
Thank you @ahuang314 for this PR! It looks good to me.

PR #338, opened 2025-06-20
Updated band requirement for velocity dispersion notebook and modified light2mass.py and test_light2mass.py to reflect changes

mia-lamontagne:
Modified get_deflector_velocity_dispersion.ipynb to run with only the three LSST bands (g, r, i) as these are the only available in the cosmo_DC2.fits catalog. 

Updated get_velocity_dispersion() in light2mass.py to support both full 5-band LSST photometry and reduced 3-band input. When only g, r, and i bands are available, the function now substitutes the missing u band with g and the z band with i, ensuring compatibility with LSST_to_SDSS(). 

Added additional tests to test_light2mass.py to reflect the new 3-band case. While error thresholds remain the same for the 3-band "weak-lensing" case, a greater error threshold of 10% was used in the "spectroscopic" case to reflect expected greater deviation from accepted value.

### DIFF
--- a/tests/test_Deflectors/test_light2mass.py
+++ b/tests/test_Deflectors/test_light2mass.py
@@ -98,6 +98,7 @@ def test_invalid_deflector_type():
             scaling_relation="spectroscopic",
         )
 
+
 def test_missing_required_bands():
     cosmo = FlatLambdaCDM(H0=70, Om0=0.3)
     deflector_type = "elliptical"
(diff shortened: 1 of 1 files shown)


sibirrer:
thank you very much @mia-lamontagne ! There is an error in the tests (click on the CI / build and scroll up), hope you can address that

sibirrer (changes requested):
Thank you @mia-lamontagne ! One more minor change and then it's ready to merge :)

nkhadka21:
Hi @mia-lamontagne , thank you very much for this PR! There is a minor suggestion from @sibirrer which you can address. Then, the PR will be ready to merge! Thank you!

PR #339, opened 2025-07-02
Make second_brightest_image cuts on multiple bands

padma18-vb:
This is a small enhancement that allows the user to filter out lensed point sources using limiting magnitudes on multiple bands.

### DIFF
--- a/slsim/lens.py
+++ b/slsim/lens.py
@@ -254,8 +254,8 @@ class Lens(LensedSystemBase):
             magnitude of the second brightest image and corresponding
             band. If provided, selects lenses where the second brightest
             image has a magnitude less than or equal to provided
-            magnitude. eg: second_bright_image_cut = {"band": "i",
-            "second_bright_mag_max": 23}
+            magnitude. eg: second_bright_image_cut = {"band": ["i"],
+            "mag_max":[23]}
         :return: A boolean or dict of boolean.
         """
         validity_results = {}
@@ -293,8 +293,8 @@ class Lens(LensedSystemBase):
             magnitude of the second brightest image and corresponding
             band. If provided, selects lenses where the second brightest
             image has a magnitude less than or equal to provided
-            magnitude. eg: second_bright_image_cut = {"band": "i",
-            "mag_max": 23}
+            magnitude. eg: second_bright_image_cut = {"band": ["i"],
+            "mag_max":[23]}
         :param source_index: index of a source in source list.
         :return: boolean
         """
@@ -371,7 +371,11 @@ class Lens(LensedSystemBase):
         # the magnitude less or equal to "second_bright_mag_max" provided in the dict
         # second_bright_image_cut.
         if second_brightest_image_cut is not None:
-            for band_max, mag_max in second_brightest_image_cut.items():
+            cut_values = list(second_brightest_image_cut.values())
+            bands = cut_values[0]
+            values = cut_values[1]
+            image_cut_dict = dict(zip(bands, values))
+            for band_max, mag_max in image_cut_dict.items():
                 if self._source_type == "extended":
                     image_magnitude_list = (
                         self.extended_source_magnitude_for_each_image(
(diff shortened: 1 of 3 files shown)


sibirrer:
Thank you @padma18-vb! I think this feature is great. Looks like some tests need to be updated as well to reflect this new convention of the settings :)

slsim/lens.py
```diff
@@ -254,8 +254,8 @@ def validity_test(
             magnitude of the second brightest image and corresponding
             band. If provided, selects lenses where the second brightest
             image has a magnitude less than or equal to provided
-            magnitude. eg: second_bright_image_cut = {"band": "i",
-            "second_bright_mag_max": 23}
+            magnitude. eg: second_bright_image_cut = {"band": ["i"],
+            "mag_max":[23]}
```
nkhadka21, line 258: The old description was incorrect. Can you make this second_bright_image_cut = {"i": 23}. This the correct example.

slsim/lens.py
```diff
@@ -293,8 +293,8 @@ def _validity_test(
             magnitude of the second brightest image and corresponding
             band. If provided, selects lenses where the second brightest
             image has a magnitude less than or equal to provided
-            magnitude. eg: second_bright_image_cut = {"band": "i",
-            "mag_max": 23}
+            magnitude. eg: second_bright_image_cut = {"band": ["i"],
```
nkhadka21, line 296: Here also, exaple should be : second_bright_image_cut = {"i": 23}.

slsim/lens.py
```diff
@@ -371,7 +371,11 @@ def _validity_test(
         # the magnitude less or equal to "second_bright_mag_max" provided in the dict
         # second_bright_image_cut.
         if second_brightest_image_cut is not None:
-            for band_max, mag_max in second_brightest_image_cut.items():
+            cut_values = list(second_brightest_image_cut.values())
+            bands = cut_values[0]
+            values = cut_values[1]
+            image_cut_dict = dict(zip(bands, values))
```
nkhadka21, line 377: I realized that currently we are demanding what you are preparing here as image_cut_dict. I think one can just pass second_brightest_image_cut = {"i": 23, "g": 24, "r": 22}. This is exactly the same form as image_cut_dict right?

slsim/lens_pop.py
```diff
@@ -140,8 +140,8 @@ def draw_population(
             cuts that one wants to apply to the lens. eg:
             kwargs_lens_cut = {}"min_image_separation": 0.5,
             "max_image_separation": 10, "mag_arc_limit": {"i", 24},
-            "second_brightest_image_cut": {"i", 24}}. all these cuts are
-            optional.
+            "second_bright_image_cut = {"band": ["i"], "mag_max":[23]}.
```
nkhadka21, line 143: The previous given example is currect. If we demand that form.

slsim/lens_pop.py
```diff
@@ -150,8 +150,8 @@ def draw_population(
             magnitude of the second brightest image and corresponding
             band. If provided, selects lenses where the second brightest
             image has a magnitude less than or equal to provided
-            magnitude. eg: second_bright_image_cut = {"band": "i",
-            "second_bright_mag_max": 23}
+            magnitude. eg: second_bright_image_cut = {"band": ["i"],
+            "mag_max":[23]}
```
nkhadka21, line 154: Here also old example is currect.

nkhadka21 (changes requested):
Hi @padma18-vb , thank you very much for this PR! I think the old version of this feature works for multi band but need to pass multi band cuts as dict. eg: second_brightest_image_cut = {"i": 23, "g": 24, "r": 22}. There are some minor inconsistency in documentation in validity function. It would be great if you could fix them.

slsim/lens.py
padma18-vb: Oh I see! I followed the old description so I thought this feature didn't exist -- will make changes to the docstring instead!

nkhadka21 (approved):
Hi @padma18-vb , thank you very much for fixing all these! It looks very good to me.

PR #340, opened 2025-07-03
Optimization to the Quasar Host-Matching Code

timedilatesme:
I noticed that the previous quasar host-matching code I wrote was taking significant amount of time to produce a large sample so I optimized it.


nkhadka21 (approved):
Hi @timedilatesme , thank you very much for this PR! It looks good to me.

PR #341, opened 2025-07-14
add functionality to use real source galaxies from cosmos catalog

ahuang314:
This PR accomplishes three things:

1. Introduced a new source type CatalogSource. For this source, a source_dict defining the angular size, n_sersic, e1, and e2 is provided. These parameters are then matched to a real source galaxy in the provided catalog, which is then used for lensing. In order to test this feature, I have added a test catalog to the repository under data/test_COSMOS_23.5_training_sample, which has a total size of 210 KB.

2. Improved the structure of the Source, ExtendedSource, PointSource, and PointPlusExtendedSource classes by changing their initialization, so that there are separate dictionaries for the PointSource and ExtendedSource classes.

3. Same as previous bullet point but for the Galaxies, PointSources, and PointPlusExtendedSources classes.

Note: This PR looks huge at first glance but it is just a bunch of very small changes across many files


slsim/Sources/SourceTypes/real_cosmos.py
```diff
@@ -0,0 +1,354 @@
+import numpy as np
```
nkhadka21, line 1: Same. If we can use Interpolated for this class, we should avoid this class. Furthermore, we might develop Interpolated class as a general class to use real galaxies? What do you think?

slsim/Sources/SourceTypes/real_cosmos.py
```diff
@@ -0,0 +1,354 @@
...
+            catalog1_path = os.path.join(cosmos_path, "real_galaxy_catalog_23.5.fits")
+            catalog2_path = os.path.join(
+                cosmos_path, "real_galaxy_catalog_23.5_fits.fits"
+            )
+            cat1 = Table.read(catalog1_path, format="fits", hdu=1)
+            cat2 = Table.read(catalog2_path, format="fits", hdu=1)
+            COSMOSSource.final_catalog = self._process_catalog(cat1, cat2)
+            COSMOSSource.cosmos_path = cosmos_path
+
+    def _process_catalog(self, cat1, cat2):
```
nkhadka21, line 43: Can you write this helper function outside of this class. Maybe somewhere in Util?

slsim/Sources/SourceTypes/real_cosmos.py
```diff
@@ -0,0 +1,354 @@
...
+            "angular_size",
+        ]
+
+        for col in filtered_catalog.colnames:
+            if col not in keep_columns:
+                filtered_catalog.remove_column(col)
+
+        return filtered_catalog
+
+    def _match_source(self):
```
nkhadka21, line 211: Is it possible to write this function outside of this class? (if you decide to use Interpolated class instead of creating this new class, same question will be for Interpolated class.)

slsim/Sources/SourceTypes/real_cosmos.py
ahuang314: The Interpolated class requires the image to be already supplied in source_dict, meaning that the user already knows which real galaxy source to use. But for this COSMOSSource class, the user does not yet know which real galaxy source to use. We only do the catalog matching when simulating an image and not during initialization. This is why I created a second class.

If we wish to keep both possibilities (one where the user already has the image of the real source, and one where the user matches parameters to a catalog), then we should have both classes. So, I disagree that we should combine the COSMOSSource and Interpolated classes. However, I agree that we can generalize the COSMOSSource class to be a CatalogSource class.

Let me know what you think. Also @sibirrer feel free to chime in.

nkhadka21: Thanks @ahuang314! Then, we can have both of these classes. As you mentioned, you can make it more general class.

nkhadka21 (approved):
Hi @ahuang314 , thank you very much for this wonderful work! It looks very good.

PR #346, opened 2025-07-30
Deflector refactoring

sibirrer:
major refactoring of the Deflector() class.
New "EPL" deflector that does not have a light component, and EPL_SERSIC that comes with a light component


nkhadka21 (approved):
Hi @sibirrer , thank you very much for this PR! It looks very good to me.

PR #347, opened 2025-07-31
minor update to microlensing notebook

timedilatesme:
I had made a figure for the SLSim paper but didn't commit to the SLSim, so here it is.


slsim/lens.py
```diff
@@ -870,6 +870,9 @@ def _point_source_magnitude_microlensing(
                 **kwargs_microlensing,
             )
         )
+
+        self.microlensing_model_class = ml_lc_lens
```
sibirrer, line 874: what does this do in the code? Neither is it documented nor has it any effect on the code

slsim/Microlensing/lightcurvelensmodel.py
```diff
@@ -343,6 +346,10 @@ def generate_magnification_maps_from_microlensing_params(
             )
             magmaps_images.append(magmap)
 
+        self.magmaps_images = (
```
sibirrer, line 349: I would make this a private variable and then a public definition to have access to it (if you want the user to) with a proper documentation

sibirrer: also, I don't see the purpose of it when you call the same definition twice, it currently re-generates this class anyways (perhaps not desired so)

sibirrer (changes requested):
Thank you! I have some minor comments

slsim/lens.py
timedilatesme: I added these recently to be able to access the `MicrolensingLightCurveFromLensModel` class directly from the lens class. I can make these private if that's better. I still need to update their documentation, I will do that!

slsim/Microlensing/lightcurvelensmodel.py
timedilatesme: Sure! will make it a private variable and provide the public definition for access! the reason for storing these maps in a variable was so that it is directly retrieved from the cache and we don't re-run the function, which may generate some new maps. will change the definition to be only self._magmaps_images in the code.

sibirrer (approved):
Looks good @timedilatesme ! FYI, I am in a major refactoring now and hopefully will be done(ish) in few days

PR #348, opened 2025-08-01
Source module refactoring

sibirrer:
main refactoring of Source module


nkhadka21 (commented):
Hi Simon, thank you very much for this refactoring! It looks very good to me. I have 2 minor comments, otherwise it is ready to merge. Also, it would be great if you could increase the codcov to the expected limit!

PR #349, opened 2025-08-06
Simulating AGN variability by generating snapshots of the variable disk

padma18-vb:
This pull request will add a flag to the AGN variability method. It will allow the user to:
1. generate snapshots of the variable disk over the observing time,
2. multiply the disk emission with the varying microlensing map over some trajectory at the position of the lensed image 
3. output AB magnitudes from the magnified disk emission. 

It will capture the varying morphology of the disk, and the microlensing time delay encoded into the light curves.

So far, I have added an empty method that will connect various util methods to generate these snapshots. 

This work primarily builds on AMOEBA by @Henry-Best-01.


timedilatesme:
Hi @padma18-vb , Thanks so much! I was also recently playing with the variable disks in amoeba and glad that you opened this PR! Once you finalize the function to generate variable disk snapshots then we can connect it to the Microlensing Module. We need to make some changes in the calculations there, but otherwise it should not take much refactoring.

PR #350, opened 2025-08-06
Modifications for different observatories plots (Galaxy-galaxy Lens)

FerroYx:
Modified euclid_roman_lsst-like.yml for plotting.
Add single band options for plot montage;
Add insertable parameters for RGB function to change the color scale of different observatories.
Add options for reading a lens list outside of plot montage. 
Add test functions for those part.

### DIFF
--- a/slsim/Plots/lens_plots.py
+++ b/slsim/Plots/lens_plots.py
@@ -30,7 +30,15 @@ class LensingPlots(object):
         self._observatory = observatory
         self._kwargs = kwargs
 
-    def rgb_image(self, lens_class, rgb_band_list, add_noise=True, minimum=None, stretch=None, Q=None):
+    def rgb_image(
+        self,
+        lens_class,
+        rgb_band_list,
+        add_noise=True,
+        minimum=None,
+        stretch=None,
+        Q=None,
+    ):
         """Method to generate a rgb-image with lupton_rgb color scale.
 
         :param lens_class: class object containing all information of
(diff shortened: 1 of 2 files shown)


FerroYx:
@sibirrer @nkhadka21 Please check! The line which is not covered is out of my modifications. 
Do you need me to write a test function for testing the roman config via plot_montage function?

data/SkyPy/euclid-roman-lsst-like.yml
```diff
@@ -1,6 +1,6 @@
 mag_lim: 30
 fsky: 0.1 deg2
-z_range: !numpy.arange [0, 5.01, 0.1]
+z_range: !numpy.arange [0, 3.01, 0.1]
```
sibirrer, line 3: remind me, z>3 is not possible because of how the filters are defined in Euclid speclite that go too blue. Perhaps make a comment here

slsim/Plots/lens_plots.py
```diff
@@ -107,20 +114,51 @@ def plot_montage(
             default is 1
         :param kwargs_lens_cut: lens selection cuts for
             Lens.validity_test() function
+        :param single_band: boolean flag, set to True to only show the
+            first band in the rgb_band_list, default is False
+        :param lens_class_list: list of lens_class objects to be
```
sibirrer, line 119: lens_class -> Lens() class objects

slsim/Plots/lens_plots.py
```diff
@@ -80,6 +81,9 @@ def rgb_image(self, lens_class, rgb_band_list, add_noise=True):
             minimum = 0
             stretch = 0.5
             Q = 8
+        minimum = kwargs.get("minimum", minimum)
```
sibirrer, line 84: I don't like extracting explicitly arguments out of a dictionary. It is unclear for the user which ones can/need to be set and/or what the default values are.
I would put minimum, stretch and Q as inputs with =None, and if None, set them according to the previous definition

sibirrer (changes requested):
Thank you @FerroYx ! It looks very good. I have one request for change.

sibirrer (approved):
Looks good, thank you @FerroYx !

PR #351, opened 2025-08-06
Deflector refactoring

sibirrer:
This PR changes the design underneath the Source() and Deflector() class.

New:
Deflector model renaming:
EPL -> EPL_SERSIC
Source type renaming:
only use 'point_source_type' or 'extended_source_type'. Everything else gets assigned


sibirrer:
@nkhadka21 can we merge this PR?

PR #353, opened 2025-08-07
add notebook comparing SLSim lenses to SLACS Dinos and SLED lenses

mia-lamontagne:
This notebook compares various parameters of simulated SLSim lenses with both real SLACS Dinos lenses and SLED lenses. These comparisons are then visualized in corner plots for analysis.


nkhadka21:
Hi @mia-lamontagne , thank you very much for this PR! The notebook looks very good. Pre-commit is failing. Would be great if you could fix them!

Issue #366, opened 2025-09-08
Decrease Unit Test Runtime using Pytest-Testmon


jacob-hjortlund:
Currently the CI pipeline takes ~30 mins to run, which is done every time something is pushed to a given PR. It is rare that a single push modifies more than a handful of functions, which means that the majority of tests being rerun are unecessary. To reduce run-time on push to a PR, I propose updating the CI pipeline / Github Action to make use of the Pytest plugin [Pytest-Testmon](https://testmon.org/).

The way it functions is that it runs all tests at the first push, creates a cache over tests run, and only reruns tests if the code they cover has been changed in a later push to the PR. This way future development wont have to wait 30 mins to see if changes pass  tests / are covered, speeding up dev turnover.

Since this would require modifying the config file for the CI Github Action, I've assigned @sibirrer and @nkhadka21, since the two of you have access to making such modifications. 😄

PR #369, opened 2025-09-17
added pip and testmon db caching for speed ups during tests

jacob-hjortlund:
PR implementing #366 with caching of testmon database as well as pip

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,48 +1,45 @@
-name: CI
+- uses: actions/checkout@v4
 
-on:
-  push:
-    branches:
-      - main
-      - cleanup
-  pull_request:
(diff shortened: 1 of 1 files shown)


jacob-hjortlund:
> Thank you very much @jacob-hjortlund ! (I am currently a bit slow due to a newborn). I have one minor request. The CI failed due to something unrelated that is not as stable (did not find a halo in the sky area). If you re-trigger the test, chances are that it passes (and need to be made more stable, but not your issue).

Thanks, I thought that was the case. I was planning on seeing if I could fix a seed to remove the flakiness of the test, but have been busy with some travel + #365. Hope everything is going well with the newborn! :smile:

sibirrer:
I reassign @nkhadka21 here to help you out

PR #375, opened 2025-10-03
Correction on Ob0 cosmological parameter for cluster lenses

furcelay:
Correction on colossus cosmological parameters when importing from astropy LambdaCDM using default Ob0 = 0. In case Ob0 = 0, it will be set to 0.04897.


sibirrer (changes requested):
Thank you every much! Can you perhaps make a test function to check that Ob0 is the value what we expect when first Ob was not initialized in astropy

PR #405, opened 2026-02-19
Temperory Branch for nonlens Generation

FerroYx:
Just combine some features from Paras code.  Won't merge to the main.


FerroYx (commented):
false_positive modification.

PR #427, opened 2026-06-04
Move Supernovae into Events source directory

Zhengjingyi0823:
The Supernovae source directory has been moved into Sources/Events/ and moves the corresponding tests into test_Events. Related import paths, docs, and notebook references are updated accordingly.


sibirrer (approved):
Looks great, thank you very much @Zhengjingyi0823 !

PR #431, opened 2026-06-10
Event-lightcone

Zhengjingyi0823:
Add event lightcone and its test


Zhengjingyi0823:
Hi @sibirrer, I updated the PR. The checks seem to be passing now. Could you review it when you have time?

slsim/Sources/Events/event_lightcone.py
```diff
@@ -0,0 +1,68 @@
+from skypy.galaxies.redshift import redshifts_from_comoving_density
+from slsim.Sources.Events.event_pop import EventPopulation
+from astropy import units
+import numpy as np
+
+
+class EventLightcone(object):
+    def __init__(self, cosmo, redshifts, sky_area, noise, time_interval, model):
+        """
```
sibirrer, line 12: can you provide a general description of the role of this class?

slsim/Sources/Events/event_lightcone.py
```diff
@@ -0,0 +1,68 @@
...
+        :param cosmo: cosmology object
+        :type cosmo: ~astropy.cosmology object
+        :param redshifts: redshifts for event density lightcone to be evaluated at
+        :type redshifts: array-like
```
sibirrer, line 17: just checking: does this need to be sorted and ascending in values?

slsim/Sources/Events/event_lightcone.py
```diff
@@ -0,0 +1,68 @@
...
+        :type model: string value
+        """
+        self._cosmo = cosmo
+        self._input_redshifts = np.asarray(redshifts)
+        self._sky_area = sky_area
+        self._noise = noise
+        self._time_interval = time_interval
+        self._model = model
+
+        event_pop = EventPopulation(self._model, self._cosmo, self._input_redshifts[-1])
```
sibirrer: perhaps make this more stable by providing specific variable names in the initialization here. i.e. EventPopulation(model=self._model....)

Also, here are self._input_redshifts[-1] asked for. I guess this is the highest redshift. This was not specified in the initialization of the class

slsim/Sources/Events/event_lightcone.py
```diff
@@ -0,0 +1,68 @@
...
+
+        # Convert source-frame event rate to observer-frame event rate
+        rate_source_frame = event_pop.event_rate(self._input_redshifts)
+        rate_observer_frame = rate_source_frame / (1 + self._input_redshifts)
+
+        self.density = self.convert_density(rate_observer_frame)
```
sibirrer: does self.density need to be accessed outside of the class? I don't think so as it is very tight to the redshift of the initialization. Make sure to keep this variable private as it is hard to guess/read its meaning from outside the class

slsim/Sources/Events/event_lightcone.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def convert_density(self, density):
+        """Converts event comoving densities from [yr^(-1)Mpc^(-3)] to have the
+        desired time unit.
```
sibirrer: what does "desired time units" mean?

slsim/Sources/Events/event_lightcone.py
```diff
@@ -0,0 +1,68 @@
...
+
+        :param density: initial comoving density of event, such as BNS
+            merger or SNIa in unit [yr^(-1)Mpc^(-3)]
+        :return: event comoving density with the desired time unit
+            [day^(-1)Mpc^(-3), hr^(-1)Mpc^(-3), etc.]
+        :return type: array-like
+        """
+        time_conversion = (1 * units.year).to(self._time_interval.unit)
```
sibirrer: I am not sure this is required here. You can also give the "density" a unit of yr-1 first. Also, why do you want to return the rate density in the input units? If I put in 365 days I will get a different output than if I would provide 1 year (because it would be converted into different units without telling me, which is dangerous. 
Why not convert any input time interval into either units of days or years to be well specified? Perhaps even keeping the units of time in there?

slsim/Sources/Events/event_lightcone.py
```diff
@@ -0,0 +1,68 @@
...
+        converted_density = density / time_conversion * self._time_interval.value
+
+        return converted_density.value
+
+    def event_sample(self):
+        """Integrates event comoving density in light cone.
+
+        :return: sampled redshifts such that the comoving number density
+            of events corresponds to the input distribution
+        :return type: numpy.ndarray
```
sibirrer, line 71: also state in what units /time this is drawn from

sibirrer (changes requested):
Thank you @Zhengjingyi0823 ! I have some comments to be more clear about units and the handling of these.

slsim/Sources/Events/event_lightcone.py
```diff
@@ -0,0 +1,79 @@
+from skypy.galaxies.redshift import redshifts_from_comoving_density
+from slsim.Sources.Events.event_pop import EventPopulation
+from astropy import units
+import numpy as np
+
+
+class EventLightcone(object):
+    """Class to integrate observer-frame event rate density per comoving volume
```
sibirrer: I don't fully understand. What does the coming volume here mean? This is the event rates within a solid angle as a function of redshift in the observer frame

slsim/Sources/Events/event_lightcone.py
```diff
@@ -0,0 +1,79 @@
...
+        :return: event rate density in [Mpc^(-3)].
+        :return type: array-like
+        """
+        time_conversion_in_years = self._time_interval.to_value(units.year)
+        converted_density = density * time_conversion_in_years
+
+        return converted_density
+
+    def event_sample(self):
+        """Integrates event comoving density in light cone.
```
sibirrer: you should describe what quantities are computed. Under the hook it does it by integrating coming event density within a light cone, but that can be changed.

What is the unit in the event rate? Is this per year? It is unclear from the description

sibirrer (changes requested):
Looks good. I have still two minor comments

sibirrer (approved):
Looks great, thank you!

PR #433, opened 2026-06-12
Slsim logo

nkhadka21:
This PR adds:
1. SLSim logo
2. Citation info


slsim/Util/slsim_logo.py
```diff
@@ -0,0 +1,434 @@
+import numpy as np
```
sibirrer, line 1: maybe this file is a bit in a too prominent location? Perhaps put it in the same folder as the logo?

notebooks/slsim_logo.ipynb
```diff
@@ -0,0 +1,84 @@
+{
```
sibirrer: do you need this notebook since it is only three lines of code? Perhaps a readme in the logo folder describing what needs to be done to create the logo?

sibirrer (changes requested):
Thank you @nkhadka21! I made some suggestion to have the code base a bit smaller

slsim/Util/slsim_logo.py
nkhadka21: If we put it outside of the slsim, we can not import it. So, I put it in the Util.

notebooks/slsim_logo.ipynb
nkhadka21: I think we don't need it. I just gave a small section in readme showing code to generate logo.

CITATION.rst
```diff
@@ -2,5 +2,8 @@ Citation Guidelines
 ===================
 
 We strive for open-source and reproducible science and all of our code is open-source available here.
-If you use part of this package for your research, please link to this repository.
-We plan to have several publications an will revise these guidelines.
\ No newline at end of file
+If you use part of this package for your research, We strive for open-source and reproducible science and all of our code is open-source available here.
+If you use part of this package for your research, please cite:
+`Khadka et al. (2026) <https://ui.adsabs.harvard.edu/abs/2026arXiv260317138K/abstract>`_.
+Additional SLSim publications may be released in the future, and these citation guidelines will
```
sibirrer, line 16: perhaps add that for the usage of individual modules, other citations may be required.  I am thinking of SkyPy, SL_hammock and other things. Just that this is not a blank check to not cite other work the code relied on

CITATION.rst
```diff
@@ -2,5 +2,8 @@ Citation Guidelines
 ===================
 
 We strive for open-source and reproducible science and all of our code is open-source available here.
-If you use part of this package for your research, please link to this repository.
-We plan to have several publications an will revise these guidelines.
\ No newline at end of file
+If you use part of this package for your research, We strive for open-source and reproducible science and all of our code is open-source available here.
```
sibirrer: "We strive" -> "we strive"

sibirrer (approved):
Thank you @nkhadka21 !
I have some comments and suggestions for the citation guidelines, but approve here already.

PR #434, opened 2026-06-15
Work on cluster lensing, LOS source galaxy rendering, generating random galaxy clusters

simonjs830:


slsim/Lenses/lens.py
```diff
@@ -1493,7 +1512,7 @@ def deflector_mass_model_lenstronomy(self, source_index=None):
             z_source=z_source,
             z_source_convention=self.max_redshift_source_class.redshift,
             use_jax=use_jax,
-            multi_plane=bool(self.multi_plane),
+            multi_plane=bool(self.multi_plane) if multi_plane is None else multi_plane,
```
sibirrer, line 1535: The code works. Perhaps it is better to see how self.multi_plane is initialized and adapt that instead of additional criteria here

slsim/Lenses/lens_pop.py
```diff
@@ -107,6 +108,163 @@ def select_lens_at_random(self, test_area=None, verbose=False, **kwargs_lens_cut
...
+        :type verbose: bool
+        :return: Lens() instance with parameters of the deflector and lens and source field.
+        :rtype: Lens
+        """
+
+        n = 0
+        while True:
+            # draw random deflector
+            _deflector = self._lens_galaxies.draw_deflector()
+            cluster_id = _deflector._deflector._deflector_dict["cluster_id"]
```
sibirrer: this line of code seems only to work for clusters and accesses a private variable. This should be avoided. Do you need that?

slsim/Lenses/lens_pop.py
```diff
@@ -107,6 +108,163 @@ def select_lens_at_random(self, test_area=None, verbose=False, **kwargs_lens_cut
...
+
+            ### compute smallest square containing all deflector subhalos, to avoid computing unnecessary caustics and source galaxies
```
sibirrer: this feature to calculate the influence area where to check for sources behind should be generalized and be part of the the Deflector() class

slsim/Lenses/lens_pop.py
```diff
@@ -107,6 +108,163 @@ def select_lens_at_random(self, test_area=None, verbose=False, **kwargs_lens_cut
...
+            bbox = [[np.inf, -np.inf], [np.inf, -np.inf]]
+            max_dist = 20
+
+            for s in _deflector._deflector._subhalos:
```
sibirrer: again, access to a private variable that may not be generally supported

slsim/Lenses/lens_pop.py
```diff
@@ -107,6 +108,163 @@ def select_lens_at_random(self, test_area=None, verbose=False, **kwargs_lens_cut
...
+            radius = max(bbox[0][1] - center[0], bbox[1][1] - center[1]) + 20
+
+            if not hasattr(self, "_cached_caustics"):
+                self._cached_caustics = {}
+
+            if cluster_id in self._cached_caustics:
+                if self._cached_caustics[cluster_id] == -1:
+                    continue
+
+                ra_caustic_list, dec_caustic_list = self._cached_caustics[cluster_id]
```
sibirrer: I see that you cache here the caustics for individual deflectors/clusters. Not sure this is needed, or you can generalize it to deflector IDs. The danger here is that this will use up a lot of memory when having many (potential) deflectors. Do you need to save the caustics or do you only need the area?

slsim/Lenses/lens_pop.py
```diff
@@ -107,6 +108,163 @@ def select_lens_at_random(self, test_area=None, verbose=False, **kwargs_lens_cut
...
+            else:
+                ### compute caustics to filter source galaxies for validity checking
+                _dummy_lens = Lens(
+                    deflector_class=_deflector,
+                    source_class=self._sources.draw_source(z_min=2),
+                    cosmo=self.cosmo,
+                    use_jax=self._use_jax,
+                )
+                lens_model_object, model_params = (
+                    _dummy_lens.deflector_mass_model_lenstronomy()
```
sibirrer: there might be a way that you don't need a dummy Lens() instance. You need the deflector model for a high-redshift source only

slsim/Lenses/lens_pop.py
```diff
@@ -107,6 +108,163 @@ def select_lens_at_random(self, test_area=None, verbose=False, **kwargs_lens_cut
...
+                )
+                lens_model_ext = LensModelExtensions(lens_model_object)
+                _, _, ra_caustic_list, dec_caustic_list = (
+                    lens_model_ext.critical_curve_caustics(
+                        model_params,
+                        center_x=center[0],
+                        center_y=center[1],
+                        compute_window=2 * radius,
+                        # compute_window = 150,
+                        grid_scale=2.0,  # 0.5, #TODO experiment with grid_scale
```
sibirrer: with a 2 arc second grid you would not see images appearing closer than 2" together. I know this is expensive and might not be needed to quantify whether there are at least two images.

slsim/Lenses/lens_pop.py
```diff
@@ -107,6 +108,163 @@ def select_lens_at_random(self, test_area=None, verbose=False, **kwargs_lens_cut
...
+            )
+            _source_cut = [
+                s
+                for s in _source
+                if s.redshift > _deflector.redshift
+                and np.min(
+                    (ra_caustic_list - s.extended_source_position[0]) ** 2
+                    + (dec_caustic_list - s.extended_source_position[1]) ** 2
+                )
+                < 5**2
```
sibirrer: I don't understand that part

slsim/Lenses/lens_pop.py
```diff
@@ -107,6 +108,163 @@ def select_lens_at_random(self, test_area=None, verbose=False, **kwargs_lens_cut
...
+            if len(_source_cut) < min_num_sources:
+                continue
+
+            # lens only with sources near caustics to speed up validity checking
+            test_lens = Lens(
+                deflector_class=_deflector,
+                source_class=_source_cut,
+                cosmo=self.cosmo,
+                use_jax=self._use_jax,
+                multi_plane="Source",
```
sibirrer, line 200: this might not be as general

sibirrer (changes requested):
Thank you @simonjs830!
I made some comments. You also see that you did not add tests. Let me know if you have questions!

.github/workflows/tests.yml
sibirrer: change it back to 12

slsim/Deflectors/deflector.py
```diff
@@ -330,3 +315,28 @@ def critical_curves_caustics_list(
         )
 
         return ra_crit_list, dec_crit_list, ra_caustic_list, dec_caustic_list
+
+def surface_brightness(ra, dec, lens_light_model_list, kwargs_lens_light_mag):
```
sibirrer, line 276: isn't there another definition somewhere else for it?

slsim/Deflectors/deflector_group.py
```diff
@@ -0,0 +1,323 @@
...
+        )
+        theta_E_infinity = lenstronomy_util.theta_E_numerical(
+            lens_mass_model_list=lens_mass_model_list,
+            kwargs_lens_mass=kwargs_lens_mass,
+            use_jax=use_jax,
+        )
+        self._theta_e_infinity = theta_E_infinity
+        return theta_E_infinity
+
+    def critical_curves_caustics_list(self, z_source, cosmo, kwargs_critical_curve_caustics=None):
```
sibirrer: perhaps make this routine more generally accessible

sibirrer (changes requested):

PR #436, opened 2026-06-17
Adding Euclid RGB Merge Method

FerroYx:
Integrated Euclid RGB image simulation functions and related test functions.

The method is mainly based on the Euclid Official Paper:
> See https://arxiv.org/pdf/2503.15324
> for specific details on the Euclid Q1 image processing steps.

For some of the methods used in the paper that are not applicable to our simulated data, e.g., arcsinh scale(500/1/0.5), the central 100×100 pixels had a mean of 0.2, rec.709 luminance weights(not in this paper, but a popular luminance weight setting). But I still offer the options for those setting in case that we could used them in the future.

The expected input order is [VIS, Y, J] (this selection is based on this paper). VIS is treated as the high-resolution luminance channel, while Y and J provide colour information and are resampled to the VIS image shape when necessary.

See the simulation results below:

Compared with Euclid Official Image:
image

Compared with Roman and Euclid:
image


slsim/ImageSimulation/euclid_image_simulation.py
```diff
@@ -0,0 +1,785 @@
+"""Utilities for displaying pre-simulated Euclid images.
```
sibirrer: perhaps you should rename this file as the images are not simulated here, but only the color scale and stitching happens here

slsim/ImageSimulation/euclid_image_simulation.py
```diff
@@ -0,0 +1,785 @@
...
+
```
sibirrer, line 2: are some of the color routines perhaps a bit more general than Euclid? Perhaps that can be a bit generalized. Perhaps not but might be worth considering

sibirrer (changes requested):
Thank you very much @FerroYx !
I made some comments. Also, are some of the definitions copied from some other source code? If so, you should reference it

slsim/ImageSimulation/euclid_image_simulation.py
FerroYx: That's right, maybe call it euclid_rgb_util.py? And move that to `Util` folder?

sibirrer: yes, that would work. Perhaps check where the other rgb routines are as well

slsim/ImageSimulation/euclid_image_simulation.py
FerroYx: Yep, I'll clean and modify the method and the default setting of rgb config. I'll remove some of the methods that are unnecessary and completely incompatible with our data.

slsim/ImageSimulation/euclid_image_simulation.py
FerroYx: I recorded those code reference at my desktop, I'll add them to the documentation.

FerroYx:
Here is the new Euclid RGB (with consistent value max & min bands)
image

sibirrer:
Great @FerroYx ! Is the PR ready to be reviewed again?

FerroYx:
@sibirrer Hi Simon! It's ready to be merged.

sibirrer (approved):
Looks good, thank you @FerroYx !

PR #438, opened 2026-06-24
Add band-dependent Double-sersic/Catalog-Source color gradients

FerroYx:
# Intro

Based on [MegaMorph](https://arxiv.org/pdf/1212.3332) we know that galaxy structural parameters and flux distribution can vary smoothly with wavelength.

And [On the shear estimation bias induced by the spatial variation of
colour across galaxy profiles](https://arxiv.org/pdf/1211.5025) displays that Spatial color gradients cause differences in the effective morphology across different bands and affect chromatic PSF and shear measurements.

So far, `Slsim`'s color gradient is a simple model fixed for any band. And it's similar for Catalog Source (HST), which only offer single band images.

To achieve color gradients that more closely reflect reality, I will update the Double Sersic and Catalog Source modules to provide band-dependent color gradients.

## Model

For a reference-band component weight `w0`, the first-component weight in band
`b` is defined as

\[
w_0(b) =
\mathrm{sigmoid}\left[
\mathrm{logit}(w_0) +
s \left(p(b) - p(b_{\rm ref})\right)
\right],
\]

where `p(b)` is the normalized position of the band in wavelength space and
`s` is `color_gradient["strength"]`. The second-component weight is
`w1(b) = 1 - w0(b)`.

This form was chosen as a phenomenological parameterization because it:

- recovers the original `w0` and `w1` in the reference band;
- keeps both component weights positive and normalized;
- preserves the total flux in every simulated band;
- provides a single signed parameter controlling whether component 0 is
  relatively redder (`strength > 0`) or bluer (`strength < 0`);
- orders LSST, Euclid, and Roman bands consistently through approximate
  central wavelengths.

This is not intended to be a physical component-SED model or a direct fit to
a particular observed colour-gradient relation. It provides a controlled
first-order model for multi-band simulation and sensitivity tests.


FerroYx:
Still in progress; currently revising the HST section and adding a fallback option to Double Service in case there is no match in the HST source.

sibirrer:
@FerroYx I am in the process of refactoring the Deflector() class to have access to all the Source() class features. So might not all be needed here for the deflectors.
It's just a bit of work and I hope I have it done by tomorrow

FerroYx:
@sibirrer No problem, I'll wait till you update that part and make necessary modification.

> @FerroYx I am in the process of refactoring the Deflector() class to have access to all the Source() class features. So might not all be needed here for the deflectors. It's just a bit of work and I hope I have it done by tomorrow

sibirrer:
@FerroYx we have merged a major PR for deflector re-factoring. So now, you only need the Source() classes to be updated to have this enabled. Let me know in case there are any issues with the merge conflict

slsim/Deflectors/deflector_util.py
```diff
@@ -121,7 +121,13 @@ def light2mass(
             )
         else:
             # scale light to mass ellipticity
-            e1_light, e2_light = kwargs_source["e1"], kwargs_source["e2"]
+            if "e1" in kwargs_source and "e2" in kwargs_source:
+                e1_light, e2_light = kwargs_source["e1"], kwargs_source["e2"]
+            else:
+                e1_light, e2_light = (
```
sibirrer: this is a bit tricky. I would have liked the total effective ellipticity to be used to connect mass and light, and not just one component (which might not be the dominating one)

slsim/ImageSimulation/image_quality_lenstronomy.py
```diff
@@ -256,3 +278,59 @@ def get_all_supported_bands():
...
+    :raises ValueError: if the band is not registered.
+    """
+    if band in _DEFAULT_BAND_CENTRAL_WAVELENGTH_MICRON:
+        return _DEFAULT_BAND_CENTRAL_WAVELENGTH_MICRON[band]
+
+    obs_name = get_observatory(band)
+    bands = _OBSERVATORY_REGISTRY[obs_name]["bands"]
+    if len(bands) == 1:
+        return 0.0
+    return float(bands.index(band)) / float(len(bands) - 1)
```
sibirrer: I don't know what is happening here. I would have assumed that you use the weighted wavelength dependent transmission

slsim/ImageSimulation/image_quality_lenstronomy.py
```diff
@@ -10,6 +10,28 @@
 LSST_BAND_LIST = ["u", "g", "r", "i", "z", "y"]
 EUCLID_BAND_LIST = ["VIS", "Y", "J", "H"]
 
+_DEFAULT_BAND_CENTRAL_WAVELENGTH_MICRON = {
```
sibirrer: this is a bit dangerous as the naming convention for the bands are not handled here.
There should be a more central part where the band configurations are handled by. Tagging here @timedilatesme that might also give you a direction

slsim/ImageSimulation/image_quality_lenstronomy.py
```diff
@@ -256,3 +278,59 @@ def get_all_supported_bands():
...
+
+    obs_name = get_observatory(band)
+    bands = _OBSERVATORY_REGISTRY[obs_name]["bands"]
+    if len(bands) == 1:
+        return 0.0
+    return float(bands.index(band)) / float(len(bands) - 1)
+
+
+def get_band_normalized_position(band, reference_band=None):
+    """Return a normalized wavelength position for a registered band.
```
sibirrer: I don't know what "position" here means.
I don't know what this definition should achieve

slsim/Sources/SourcePopulation/galaxies.py
```diff
@@ -428,12 +452,37 @@ def convert_catalog_to_source(
...
+                    radius_factors[0]
+                )
+                kwargs_source["angular_size_1"] = angular_size * float(
+                    radius_factors[1]
+                )
             else:
                 raise ValueError(
-                    "semi-major and semi-minor axis are missing for the first light"
-                    " profile in galaxy_list columns %s" % colnames
+                    "semi-major and semi-minor axis or angular_size are missing "
```
sibirrer: this is a less specific error message. Make it clear what the issue is, not and/or

slsim/Sources/SourceTypes/catalog_source.py
```diff
@@ -93,6 +116,24 @@ def __init__(
                 f"Catalog_type {catalog_type} not supported. Currently only {CATALOG_TYPES} are supported."
             )
 
+        if self._band_dependent_color_gradient:
+            if catalog_type != "HST_COSMOS":
+                raise ValueError(
+                    "band_dependent_color_gradient is currently supported only "
+                    "for HST_COSMOS."
```
sibirrer: make sure this is for catalog_type "HST_COSMOS". You can also print the actual catalog type used for better readability

slsim/Sources/SourceTypes/catalog_source.py
```diff
@@ -93,6 +116,24 @@ def __init__(
...
+                )
+            if not isinstance(self._color_gradient, dict):
+                raise ValueError(
+                    "color_gradient must be a dictionary when "
```
sibirrer, line 125: add in error message what color_gradient actually is at input

sibirrer (changes requested):
Thank you very much! I give here a first batch of comments

slsim/Sources/SourceTypes/catalog_source.py
FerroYx: Updated
`raise ValueError(
    "band_dependent_color_gradient is currently supported only "
    "for catalog_type='HST_COSMOS'; received "
    f"catalog_type={catalog_type!r}."
)`

slsim/ImageSimulation/image_quality_lenstronomy.py
FerroYx: This part has been removed. For filters available in speclite, the chromatic scaling now uses the throughput-weighted effective wavelength from the registered filter response. Roman bands, whose response curves are not distributed with speclite, use explicitly configured effective wavelengths in the observatory registry `register_observatory()`.

slsim/ImageSimulation/image_quality_lenstronomy.py
FerroYx: Same as above, I read them directly from speclite or from `register_observatory()`.

slsim/ImageSimulation/image_quality_lenstronomy.py
FerroYx: The normalized band “position” is no longer used by the color-gradient implementation. It has been replaced with the explicitly defined logarithmic effective-wavelength ratio, `log(lambda_eff,band / lambda_eff,reference)`.

This quantity is zero in the reference band, 
negative for bluer bands, 
and positive for redder bands, 
and does not depend on the number or ordering of registered filters.

slsim/Sources/SourcePopulation/galaxies.py
FerroYx: Updated.
`
raise ValueError(
    "Cannot determine DoubleSersic component sizes. Provide one of: "
    "(1) both 'angular_size_0' and 'angular_size_1'; "
    "(2) all four axis fields 'a0', 'b0', 'a1', and 'b1'; "
    "(3) a shared 'angular_size'; or "
    "(4) a catalog_type from which the shared size can be inferred. "
    f"Available galaxy_list columns are {colnames}."
)
`

slsim/Lenses/lens.py
```diff
@@ -1381,15 +1381,20 @@ def lenstronomy_kwargs(
             lens_light_model_list,
             kwargs_lens_light,
         ) = self.deflector.light_model_lenstronomy(band=band)
+        # Some light classes expose their internal model list directly.  Work
+        # with copies here so adding field galaxies does not mutate the
+        # deflector and accumulate models across repeated/band-dependent calls.
+        lens_light_model_list = list(lens_light_model_list)
```
sibirrer: can we instead make deflector.light_model_lenstronomy() always return a list for the lens_light_model_list?

slsim/Lenses/lens.py
```diff
@@ -1381,15 +1381,20 @@ def lenstronomy_kwargs(
...
+        kwargs_lens_light = list(kwargs_lens_light)
```
sibirrer: dito to comment above

slsim/Util/color_gradient.py
```diff
@@ -0,0 +1,196 @@
+import numpy as np
+
+
+def default_reference_band(source_dict, default="i"):
+    """Choose the available source band closest to a default reference band."""
```
sibirrer: can you do a better documentation of this definition?

sibirrer (changes requested):
Thank you very much @FerroYx! I have some minor comments left that might help the code structure

slsim/Lenses/lens.py
FerroYx: Updated, now the return of the `light_model_lenstronomy()` are lists directly!

FerroYx:
@sibirrer Please check again

slsim/Deflectors/deflector.py
```diff
@@ -203,7 +203,10 @@ def light_model_lenstronomy(self, band=None):
         :type band: str
         :return: lens_light_model_list, kwargs_lens_light
         """
-        return self.light.kwargs_extended_light(band=band)
+        light_model_list, kwargs_light_list = self.light.kwargs_extended_light(
```
sibirrer, line 206: which ones of the light models did not return a list? I'd rather you fix it at the source then with an additional condition in the code

sibirrer (changes requested):

slsim/Deflectors/deflector.py
FerroYx: All light models already return lists. The issue was specifically that DoubleSersic.kwargs_extended_light() returned its internal _light_model_list directly, which was later mutated by extend() when field galaxies were added. I fixed this at the source by returning a copy from DoubleSersic and removed the conversion from Deflector.light_model_lenstronomy().

FerroYx: @sibirrer Please check again!

sibirrer (approved):
Thank you very much @FerroYx !

PR #440, opened 2026-06-25
Update notebooks to use EventPopulation

Zhengjingyi0823:
This PR updates the overview notebooks for event populations and replace sn lightcone with event lightcone.


Zhengjingyi0823:
One possible follow-up improvement I noticed is that the redshift-time interpolation shown in the event population notebook should not be model-dependent. It should only depend on the cosmology and redshift range, although it is currently demonstrated separately for the BNS merger and Type Ia supernova examples.

Would it make sense to add a few lines to the EventPopulation class to expose z_from_time there? Then the notebook could plot the redshift-time relation directly through EventPopulation, instead of calling the model-specific BNS or SNIa classes for this part. I kept the current structure for now, but I would like to update it later if this organization seems better.

sibirrer:
> Would it make sense to add a few lines to the EventPopulation class to expose z_from_time there? Then the notebook could plot the redshift-time relation directly through EventPopulation, instead of calling the model-specific BNS or SNIa classes for this part. I kept the current structure for now, but I would like to update it later if this organization seems better.

Yes, the z_from_time should not be specific to BNS and/or SNIa. It can also be a joint inheritance. Certainly no need to introduce it twice

sibirrer (approved):
Thank you @Zhengjingyi0823 !

PR #443, opened 2026-07-03
minor refactoring to allow precomputing magmaps and pickle lens objects

timedilatesme:


timedilatesme:
Hello @sibirrer ! This PR is ready for review. There are two lines in magmap.py which cannot be tested without generating the mag maps with GPU.

Note: due to IPM dependency on C libraries pickling lens class was not possible unless we delete the IPM object.

sibirrer (approved):
Looks good @timedilatesme .
I am wondering whether some of the micro-lensing related parts can be a bit outsourced from the Lens() class as the Lens() class gets larger and larger with 1000+ lines.
Can be done with a different PR

timedilatesme:
Sounds good! Yeah will refactor lens.py to shift these things.

PR #445, opened 2026-07-09
Add kilonova event source type

Zhengjingyi0823:
This PR adds a Redback-based Kilonova light-curve wrapper and integrates it as a new KilonovaEvent source type. It also registers the new "kilonova" keyword in source.py and adds tests for the kilonova model and event-level behavior.


Zhengjingyi0823:
The remaining CI failures seem unrelated to the kilonova changes. The failing tests are in the lens/profile and plotting parts.

I also added redback to requirements.txt because the CI environment did not have it installed, and the new kilonova model depends on it. One concern is that kilonova is an optional source type, so I wonder whether redback should be handled as an optional dependency instead of being required for all SLSim users. Please let me know which approach you prefer.

sibirrer:
Yes, I think I can fix the CI with my big change all at once

requirements.txt
```diff
@@ -10,6 +10,8 @@ colossus
 speclite
 pyyaml
 matplotlib
+redback>=1.17.0
+colossus
```
sibirrer:
```suggestion
```

requirements.txt
```diff
@@ -10,6 +10,8 @@ colossus
 speclite
 pyyaml
 matplotlib
+redback>=1.17.0
```
sibirrer: this is a very specific package only for kilonovae. Might be better to have it only in the test_requirements.txt

slsim/Sources/Events/BNSMerger/kilonova.py
```diff
@@ -0,0 +1,133 @@
...
+        modeldir=None,
+        **kwargs,
+    ):
+        """
+        :param redshift: The redshift of the kilonova source.
+        :type redshift: float
+        :param model_name: The kilonova light curve model to be used. If not provided,
+            the default model is the MOSFiT-based kilonova model.
+        :type model_name: str
+        :param ejecta_mass: Ejecta masses for the kilonova components.
```
sibirrer: can you provide the units?

slsim/Sources/Events/BNSMerger/kilonova.py
```diff
@@ -0,0 +1,133 @@
...
+        :type ejecta_mass: array-like or None
+        :param ejecta_velocity: Ejecta velocities for the kilonova components.
```
sibirrer: units?

slsim/Sources/Events/BNSMerger/kilonova.py
```diff
@@ -0,0 +1,133 @@
...
+        :type ejecta_velocity: array-like or None
+        :param opacity: Opacities for the kilonova components.
+        :type opacity: array-like or None
+        :param temperature_floor: Temperature floors for the kilonova components.
```
sibirrer: units (K?)

slsim/Sources/Events/BNSMerger/kilonova.py
```diff
@@ -0,0 +1,133 @@
...
+            "ejecta_mass": ejecta_mass,
+            "ejecta_velocity": ejecta_velocity,
+            "opacity": opacity,
+            "temperature_floor": temperature_floor,
+        }
+
+        for name, values in parameter_groups.items():
+            if values is None:
+                raise ValueError(f"{name} must be provided.")
+            if len(values) != 3:
```
sibirrer: why is this test here? Why are there three values? Can you explain what you test here in the code?

slsim/Sources/SourceTypes/kilonova_event.py
```diff
@@ -0,0 +1,145 @@
...
+        )
+        self.name = "BNS"
+        self._variability_computed = False
+        self._kwargs_variability = kwargs_variability
+        self._lightcurve_time = lightcurve_time
+
+        self._model_name = model_name
+        self._mag_zpsys = mag_zpsys
+        self._modeldir = modeldir
+        self._kwargs_kilonova = kwargs_kilonova or {}
```
sibirrer: is this proper python or what does the 'or' do here?

slsim/Sources/SourceTypes/kilonova_event.py
```diff
@@ -0,0 +1,145 @@
...
+
+            supported_bands = get_all_supported_bands()
+            provided_bands = set(supported_bands) & set(self._kwargs_variability)
+
+            for element in provided_bands:
+                name = "ps_mag_" + element
+                times = self._lightcurve_time
+
+                # Use the sncosmo band-name mapping since Redback expects registered filter names.
+                provided_band = get_sncosmo_filtername(element)
```
sibirrer, line 106: do we need SNCosmo band names? don't understand how this is required

sibirrer (changes requested):
Thank you very much!
I could resolve the merge conflict with the new refactoring. I have some minor request before the merge

slsim/Sources/Events/BNSMerger/kilonova.py
```diff
@@ -0,0 +1,136 @@
...
+                # The MOSFiT kilonova model uses three ejecta components.
+                # Each of these parameters must therefore contain three values.
+                raise ValueError(f"{name} must have three components.")
+
+        self._model_parameters = {
+            "mej_1": ejecta_mass[0],
```
sibirrer: perhaps have all these parameters as inputs to this class instead of re-naming them

sibirrer:
Thank you @Zhengjingyi0823 ! The CI fails because some tests were not updated

sibirrer:
From a quick search, perhaps add in the test files where the plots are generated these lines of code:
import matplotlib.pyplot as plt

plt.rcParams['text.usetex'] = False

Zhengjingyi0823:
I added the suggested Matplotlib setting to the plot test file. However, GitHub Actions seems to be experiencing an outage. I tried pushing several times today, but the checks were either stuck in the queue or cancelled. Should I try rerunning them tomorrow?

requirements.txt
```diff
@@ -9,7 +9,7 @@ scipy
 colossus
 speclite
 pyyaml
-matplotlib
+matplotlib==3.11.1
```
sibirrer: do you still need that with the additional fix of not using latex?
I also added trough a separate PR #466 the pit.rcparams parts. Surprised it did not result in a merge conflict or pre-commit cit.

sibirrer (commented):
Yeah, the GitHub action is either down or we are out of credits. Perhaps we wait and try tomorrow again to merge it

sibirrer:
Ok, looks good now. Just one test that through Source() you can enable the BNS class needs to be tested (two missing lines in the tests) - thank you!

sibirrer (approved):
Looks great @Zhengjingyi0823 ! Let me know if you want me to merge it

PR #446, opened 2026-07-09
Deflector refactoring

sibirrer:


slsim/Deflectors/deflector_util.py
```diff
@@ -0,0 +1,194 @@
...
+        if "e1_mass" in halo_columns and "e2_mass" in halo_columns:
+            e1_mass, e2_mass = float(halo_dict["e1_mass"]), float(halo_dict["e2_mass"])
+        elif "e_h" in halo_columns and "p_h" in halo_columns:
+            # SL_hammock conventions
+            e1_mass, e2_mass = galaxy_projected_eccentricity(
+                float(halo_dict["e_h"]), rotation_angle=np.deg2rad(halo_dict["p_h"])
+            )
+        else:
+            # scale light to mass ellipticity
+            e1_light, e2_light = kwargs_source["e1"], kwargs_source["e1"]
```
timedilatesme: I think this is a bug, and can be changed to:

```
e1_light, e2_light = kwargs_source["e1"], kwargs_source["e2"]
```

slsim/Deflectors/MassTypes/mass_base.py
```diff
@@ -0,0 +1,75 @@
...
+        """
+        pass
+
+    @property
+    def mass_ellipticity(self):
+        """Mass ellipticity.
+
+        :return: e1_mass, e2_mass
+        """
+        e1_mass, e2_mass = float(self._deflector_dict["e1_mass"]), float(
```
timedilatesme: Just checking: I think the _deflector_dict is not set in the child classes? so perhaps we can use self._e1_mass and self._e2_mass instead?

I think this method is just duplicating the ellipticity method in child classes?

slsim/Deflectors/deflector.py
```diff
@@ -76,41 +95,61 @@ def velocity_dispersion(self, cosmo=None):
...
+            reference from where within a circle the source position is
+            being drawn from
+        :type reference_position: 2d numpy array
+        :param area: area (in solid angle arc-seconds^2) to dither the
+            center of the source
+        :param center_x: RA position [arc-seconds] (optional, otherwise
+            renders within area)
+        :param center_y: DEC position [arc-seconds] (optional, otherwise
+            renders within area)
+        :return: Source() instance updated with new center position
```
timedilatesme, line 124: Now that Sources and Deflectors are on same footing so keeping the docs with Source() is probably fine? But I think since this is specifically in deflector class so we can keep it as "Deflector instance". Just a thought

slsim/Deflectors/deflector.py
```diff
@@ -140,17 +179,21 @@ def mass_ellipticity(self):
...
+    def mass_model_lenstronomy(self, lens_cosmo, spherical=False):
         """Returns lens model instance and parameters in lenstronomy
         conventions.
 
         :param lens_cosmo: lens cosmology model
         :type lens_cosmo: ~lenstronomy.Cosmo.LensCosmo instance
         :return: lens_mass_model_list, kwargs_lens_mass
         """
-        return self._deflector.mass_model_lenstronomy(lens_cosmo=lens_cosmo)
+        if lens_cosmo.z_lens >= lens_cosmo.z_source:
```
timedilatesme, line 192: This no lensing condition line is also handled in the mass_model_lenstronomy method inside the epl.py ([line 99](https://github.com/LSST-strong-lensing/slsim/pull/446/changes#diff-21567fd1b93e5d0e57f9119f21a51c3aea3417a739e1d010c95f5016c4ecd8a9L84)). There it is handled differently by setting theta_E = 0. But I don't see it for other mass profiles so perhaps we can keep it here only and remove from epl.py? 

Please disregard this comment if the conditional check in epl.py was intentional.

slsim/Deflectors/MassTypes/epl.py
timedilatesme, line 99: In the method argument we have as default lens_cosmo = None, I think we always have to provide a lens_cosmo instance otherwise this line will crash. So we make it a required non-optional argument?

slsim/Deflectors/MassTypes/epl.py
```diff
@@ -56,18 +61,28 @@ def _einstein_radius(self, lens_cosmo=None):
...
-                kappa_ext=0,
-                sis_convention=self._sis_convention,
-            )
-            return theta_E
+            if (
+                self._gamma_pl == 2
+                or self._sis_convention is True
+                or self._light.extended_source_type is None
+            ):
+                theta_E = lens_cosmo.sis_sigma_v2theta_E(self.velocity_dispersion())
```
timedilatesme, line 69: In the method argument we have as default lens_cosmo = None, I think we always have to provide a lens_cosmo instance otherwise this line will crash. So we make it a required non-optional argument?

slsim/Deflectors/MassTypes/nfw.py
```diff
@@ -0,0 +1,99 @@
+from slsim.Util.param_util import ellipticity_slsim_to_lenstronomy
+from slsim.Deflectors.MassTypes.mass_base import MassBase
+from slsim.Deflectors.MassLightConnection.velocity_dispersion import vel_disp_nfw
+
+
+class NFW(MassBase):
+    """Class of a NFW+Hernquist lens model with a Hernquist light mode."""
```
timedilatesme: I think this is a typo? as this is just the NFW model and not hernquist.

slsim/Sources/source.py
timedilatesme, line 282: just checking: this is supplied in observer frame? in the point_source_magnitude method we are using the source frame time now to supply to the self._source class. So in the line below self._source.kwargs_point_source should also receive the time in source rest frame?


In an upcoming PR, I will make all the source classes (accessed via self._source) to accept time in source rest frame.

slsim/Sources/SourcePopulation/galaxies.py
```diff
@@ -340,85 +255,251 @@ def galaxy_projected_eccentricity(ellipticity, rotation_angle=None):
...
+    include_all_keywords=False,
+):
+    """Converts input table entries into the quantities used in slsim Source()
+    class.
+
+    :param galaxy: dictionary or table entry
+    :param include_all_keywords: if True, includes all keywords and not
+        just the ones required
+    :type include_all_keywords: bool
+    :param galaxy: entry in galaxy_list
```
timedilatesme: "galaxy" is repeated here. Also a few other parameters are not documented. I am happy to fix these in a later PR as these are just minor things.

timedilatesme (approved):
Hello @sibirrer ! Thank you so much for this great PR!!! and sorry for slightly delayed review. I tried to go through as much as I understood and identified a few bugs/modifications that might make it better. Also, happy to fix some of the things myself in a later PR as well! 

Also, just checking -- do we directly wanna merge in the main branch now or first in the deflector_refactoring branch?

slsim/Deflectors/deflector_util.py
sibirrer: ah, glad you caught this!

slsim/Deflectors/MassTypes/mass_base.py
sibirrer: yes, it's duplicate and will remove it

sibirrer:
Thank you @timedilatesme !
I have incorporated your changes and merged it into the deflector_refactoring branch of the main repo. If you could update the Sources parts and the notebooks to merge it into main, that would be great. Don't try to do too many things here as I won't be able to review your changes (too many files and lines already in there) and afterwards make new smaller pull requests - thank you!

PR #448, opened 2026-07-14
Deflector refactoring

sibirrer:
Major refactoring of the Deflector() class and many other small design changes


sibirrer:
@timedilatesme are you good to merge it such that we can start the hack on this basis? We can clean up the notebook today with the help of @nkhadka21 and @huangxy256

PR #450, opened 2026-07-15
minor fixes to make source classes use rest frame time

timedilatesme:


slsim/Sources/SourceTypes/supernova_event.py
```diff
@@ -112,11 +118,13 @@ def light_curve(self):
 
                 name = "ps_mag_" + element
                 times = self._lightcurve_time
+                # sncosmo expects observer-frame times, so we convert to observer-frame times using the redshift
+                observer_frame_times = times * (1 + self._z)
```
sibirrer: do you need to do this in the for-loop?

sibirrer (approved):
Thank you for the necessary changes and clarifying documentation. To confirm, the prior use with SNCosmo was correct and this PR makes it identical back to what it was?
I have one minor suggestion but approve the PR already

timedilatesme:
> Thank you for the necessary changes and clarifying documentation. To confirm, the prior use with SNCosmo was correct and this PR makes it identical back to what it was? I have one minor suggestion but approve the PR already

Yes, previously SNe implementation was correct and now it is identical to what it was! and previously everything was supplied in the observer frame. But the AGN were not and now I have verified that it is also correct after we removed the (1+z) factor in previous PR. 

image

slsim/Sources/SourceTypes/supernova_event.py
timedilatesme: fixed this now! will merge after the tests complete.

PR #451, opened 2026-07-17
re ran the notebooks and made some small fixes

timedilatesme:


sibirrer:
Thank you @timedilatesme ! there are some minor pre-commit complaints but other than that, looks great!

slsim/Sources/SourceTypes/quasar.py
```diff
@@ -55,6 +55,7 @@ def __init__(
         :param kwargs_variability_model: Pre-computed variabilities for each band (default=None)
         """
 
+        kwargs.setdefault("allow_more_source_dict", True)
```
sibirrer, line 58: for now, this is ok, but we should change that moving forward

sibirrer (approved):

sibirrer:
no need to wait for tests as this is an interim-PR

PR #452, opened 2026-07-21
added Lens population from catalog

sibirrer:
made import statements conditional to option chosen moved LensPopulation to new subfolder


sibirrer:
@padma18-vb here I have a small skeleton class that can operate on a single Table for deflector and source parameters

slsim/Lenses/LensPopulation/lenses_from_catalog.py
```diff
@@ -0,0 +1,116 @@
...
+from slsim.Deflectors.deflector_util import deflector_from_table
+from slsim.Lenses.lens import Lens
+from slsim.Sources.source import Source
+from slsim.LOS.los_individual import LOSIndividual
+
+
+class LensPopCatalog(object):
+    """Class to deal with pre-established set of lenses.
+
+    The catalog shoudl contain parameters for the deflector ending with
```
padma18-vb: should

slsim/Lenses/LensPopulation/lenses_from_catalog.py
```diff
@@ -0,0 +1,116 @@
...
+        deflector_dict, source_dict, los_dict = _catalog_deflector_source_split(
+            lens_object
+        )
+        deflector = deflector_from_table(
+            deflector_dict,
+            mass_type=self._deflector_mass_type,
+            extended_source_type=self._deflector_light_type,
+            cosmo=self._cosmo,
+        )
+        source = Source(
```
padma18-vb, line 88: I think having the required and optional inputs for every class, and their meaning recorded somewhere in the readthedocs would be very helpful.

slsim/Lenses/LensPopulation/lenses_from_catalog.py
```diff
@@ -0,0 +1,116 @@
...
+            extended_source_type=self._extended_source_type,
+            point_source_type=self._point_source_type,
+            **source_dict
+        )
```
padma18-vb, line 92: I think it's a little confusing that the way we create a Deflector and a Source is different? Should we make it consistent?

slsim/Lenses/LensPopulation/lenses_from_catalog.py
```diff
@@ -0,0 +1,116 @@
...
+
+        :param index:
+        :return: Lens() class
+        """
+        if index is None:
+            print(self._num_lenses, "test num lenses")
+            index = random.randint(0, self._num_lenses - 1)
+        lens_object = self._lens_catalog[index]
+
+        deflector_dict, source_dict, los_dict = _catalog_deflector_source_split(
```
padma18-vb, line 71: How do we handle point sources here? I.e if a user wants to pass in point source parameters for a quasar/SNe, do they have to pass in parameters that randomly generate a quasar or SNe?

slsim/Lenses/LensPopulation/lenses_from_catalog.py
```diff
@@ -0,0 +1,116 @@
...
+        catalog_type=None,
+        use_jax=True,
+        deflector_mass_type="EPL",
+        deflector_light_type="single_sersic",
+        extended_source_type="single_sersic",
+        point_source_type=None,
+    ):
+        """
+
+        :param lens_catalog: catalog with all deflector and source parameters with <_source>, <_deflector> and <_los>
```
padma18-vb, line 37: I'm also partial to referring to parameters by "object type_component_parameter", so `deflector_mass_e1`, `source_light_e1`, etc., and SLSim refers to some parameters this way to (e.g. `deflector_stellar_mass`), so it'd be good to keep it consistent!

slsim/Lenses/LensPopulation/lenses_from_catalog.py
```diff
@@ -0,0 +1,116 @@
...
+    """
+    if isinstance(lens_object, dict):
+        colnames = list(lens_object.keys())
+    else:
+        colnames = lens_object.colnames
+
+    deflector_dict = {}
+    source_dict = {}
+    los_dict = {}
+    for item in colnames:
```
padma18-vb, line 142: And corresponding to my [previous comment](https://github.com/LSST-strong-lensing/slsim/pull/452/changes#r3631459464), I think the change here can just be to change it to `item.startswith(...)`

sibirrer:
@padma18-vb I saw a comment from you that disappeared. The reason for deflector and sources not being identical is that deflectors need additional mass (or light-to-mass) parameters. The light component should have been done exactly the same way

sibirrer:
@padma18-vb can we merge it?

padma18-vb:
> @padma18-vb can we merge it?

Hi @sibirrer, yes I left some other comments (especially the one about point sources), but I think this can also be addressed in the future? Looks good overall, thank you!
image

padma18-vb (approved):
Just minor comments, I think my pressing question is about point sources (if we need the parameters + kwargs_variability etc) but if this is addressed in a future PR, that's okay as well! Thank you for the quick PR!

sibirrer:
Thank you @padma18-vb! Somehow these comments did not show up but are visible now

sibirrer:
> > @padma18-vb can we merge it?
> 
> Hi @sibirrer, yes I left some other comments (especially the one about point sources), but I think this can also be addressed in the future? Looks good overall, thank you! 
Point source arguments can be put in the same way. The split between deflector and source does not require changes in how point sources are treated.
We will follow up with better documentation of point sources (especially quasars)

slsim/Lenses/LensPopulation/lenses_from_catalog.py
sibirrer: Yes, I think this is more a think in the Source() class. There are many options there and so you have to dig a bit through the classes. The quasar class needs to be better documented and I think @timedilatesme and @Henry-Best-01 are on it

Issue #453, opened 2026-07-22
This notebook ERL_lensed_SNe_sims.ipynb crashes my web-browser


sylvielsstfr:
This notebook crash https://github.com/LSST-strong-lensing/slsim/blob/main/notebooks/MainNotebooks/ERL_lensed_SNe_sims.ipynb. (On my mac)

The crash occurs just at the cell below ## 2. SNe Population Generation.

Here:

```
# This file can be downloaded from here: https://github.com/LSST-strong-lensing/data_public .
# Please download this file and replace the path with your path.
#sn_modeldir = os.path.join(os.path.split(path)[0], "data/SALT3.NIR_WAVEEXT")
sn_modeldir = "/Users/dagoret/MacOSX/GitHub/LSST/2026/data_public/sncosmo_sn_models/SALT3.NIR_WAVEEXT"

supernovae_host_population = PointPlusExtendedSources(
    point_plus_extended_sources_list=supernovae_data,
    cosmo=cosmo,
    sky_area=sky_area,
    kwargs_cut={"z_min": 0.1, "z_max": 5.0},
    catalog_type="skypy",
    point_source_type="supernova",
    extended_source_type="single_sersic",
    joint_point_source_kwargs={
        "variability_model": "light_curve",
        "kwargs_variability": ["supernovae_lightcurve"] + all_sn_bands,
        "sn_type": "Ia",
        "sn_absolute_mag_band": "bessellb",
        "sn_absolute_zpsys": "ab",
        "lightcurve_time": sn_time_range,
        "sn_modeldir": sn_modeldir,
    },
)
```
I was carefull to download https://github.com/LSST-strong-lensing/data_public

I suspect a problem in sncosmo. Perhaps SALT3.NIR_WAVEEXT are not well red. Check if the files exists to avoid the crash:

```
[/Users/dagoret/MacOSX/GitHub/LSST/2026/slsim/slsim/Sources/Events/Supernovae/supernova.py:114](http://localhost:8889/Users/dagoret/MacOSX/GitHub/LSST/2026/slsim/slsim/Sources/Events/Supernovae/supernova.py#line=113): UserWarning: bandpass 'lsstu' [3105, .., 4086] outside spectral range [5167.67, .., 51676.7]
Ignoring bandpass for now. Use extended wavelength SN models found here: https://github.com/LSST-strong-lensing/data_public/tree/main/sncosmo_sn_models
  warn(
[...]

PR #462, opened 2026-07-30
making the SNe catalog generation faster and fixing warnings

timedilatesme:
This PR addresses the Issue #453  and also speeds up the SNe catalog generation and host matching.

### DIFF
--- a/slsim/Sources/Events/Supernovae/supernova.py
+++ b/slsim/Sources/Events/Supernovae/supernova.py
@@ -1,9 +1,41 @@
+import copy
 import os
 from warnings import warn
 import sncosmo
 from sncosmo.bandpasses import get_bandpass
 import numpy as np
 from astropy import cosmology
+
+# Reading a source from disk (~0.7 s for SALT3) dominates the cost of building a
+# Supernova and depends only on the model, not on the individual supernova, so a
+# population loop can pay it once instead of once per object.
+_SOURCE_CACHE = {}
+
+
+def _source_from_modeldir(modeldir, sn_type, source):
+    """Build an sncosmo source from a model directory, reusing a cached
+    template when the same model has already been read from disk.
+
+    :param modeldir: directory including files for supernova models
+    :type modeldir: str
+    :param sn_type: Supernova type (Ia, Ib, Ic, IIP, etc.)
+    :type sn_type: str
+    :param source: name of the SED model, ignored for type Ia
+    :type source: str
+    :return: `~sncosmo.Source`, independent of the cached template
+    """
+    key = (modeldir, sn_type, source)
+    if key not in _SOURCE_CACHE:
+        if sn_type == "Ia":
+            _SOURCE_CACHE[key] = sncosmo.SALT3Source(modeldir=modeldir)
+        else:
+            sed_file = os.path.join(modeldir, sn_type, source) + ".SED"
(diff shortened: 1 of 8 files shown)


timedilatesme:
Hello @sibirrer ! This PR is ready for review. I found that some of the lines in slsim/Sources/Events/Supernovae/supernova.py were not covered earlier by tests but that's because they require sncosmo model data which we don't store on the repo so that's why they are still missing coverage.

slsim/Sources/Events/Supernovae/supernova.py
```diff
@@ -1,10 +1,42 @@
...
+        if sn_type == "Ia":
+            _SOURCE_CACHE[key] = sncosmo.SALT3Source(modeldir=modeldir)
+        else:
+            sed_file = os.path.join(modeldir, sn_type, source) + ".SED"
+            phase, wave, flux = sncosmo.read_griddata_ascii(sed_file)
+            _SOURCE_CACHE[key] = sncosmo.TimeSeriesSource(
+                phase=phase, wave=wave, flux=flux
+            )
+    # Copy so that each supernova owns its own x0, x1 and c.
+    return copy.deepcopy(_SOURCE_CACHE[key])
```
sibirrer: just checking: since every Source() object gets its own SALT3 model, might this run into memory issues when creating a population of lensed SNe? Or are only the interpolated values retained? (I think so but want to double-check).

slsim/Sources/Events/Supernovae/supernovae_host_match.py
```diff
@@ -34,56 +32,41 @@ def match(self):
...
+                    candidates, p=candidate_weights / candidate_weights.sum()
+                )
             )
 
-            # Append host galaxy to the matched catalog.
-            matched_catalog.add_row(host_galaxy[0])
+        # Index the catalog once, instead of growing a table row by row.
+        matched_catalog = self.galaxy_catalog[matched_indices]
+        matched_catalog["physical_size"].unit = u.kpc
+        matched_catalog["angular_size"].unit = u.rad
```
sibirrer: are you sure the matched catalog is in radians? I might have changed the skypy convention for it. Worth checking

sibirrer (commented):
Thank you very much @timedilatesme! I have two comments worth checking. Otherwise looks good

slsim/Sources/Events/Supernovae/supernovae_host_match.py
timedilatesme: Thanks for catching this! Yeah it should be in arcsec based on how skypy convention handles it now so I have made this fix.

slsim/Sources/Events/Supernovae/supernova.py
timedilatesme: Last time I checked running the population it took ~ 5mins for 1 sq. deg. SNe Ia population. With deepcopy every supernova gets a full independent copy of the entire SALT3 model. But without deepcopy only the interpolated values will be retained. I agree that not using the deepcopy is a better idea.

I checked that if we don't use deepcopy then sncosmo internally copies the parameters of the source it is given and the templates/models are shared among the population, so supernovae built from one cached source still have their own parameters.

I have updated it to not do deepcopy now and now it takes ~3 mins for generating the SNe catalog with host matching for 1 sq. deg.

sibirrer (approved):
Thank you very much @timedilatesme !

PR #463, opened 2026-08-03
pass source cut kwargs to PointSources class

huangxy256:


sibirrer (approved):
Thank you very much @huangxy256 !

PR #469, opened 2026-08-20
minor fix to support latest numpy

timedilatesme:
With numpy 2.5.1 I was experiencing issues but I see that in requirements.txt the numpy version is set to < 2.4.0 so perhaps this is not needed right now but flagging this here just in case. Will add a fallback for backwards compatibility with numpy < 2.4

This PR also makes the `sis_convention=True` as the default option and allows the user to modify it if they want to via the `GalaxyDeflectors` class.

Have also flagged that abundance matching needs to be updated such that it is pl-slope aware in addition to M* and sigma*

### DIFF
--- a/slsim/Deflectors/DeflectorPopulation/deflectors_base.py
+++ b/slsim/Deflectors/DeflectorPopulation/deflectors_base.py
@@ -23,6 +23,7 @@ class DeflectorsBase(Galaxies):
         light_type="single_sersic",
         kwargs_mass2light=None,
         catalog_type=None,
+        sis_convention=True,
     ):
         """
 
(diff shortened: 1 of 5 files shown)


sibirrer:
Just checking @timedilatesme if this is still work in progress

timedilatesme:
> Just checking @timedilatesme if this is still work in progress

Hello @sibirrer ! Yes, working on finish it today along with all my the other PRs that are floating around.

timedilatesme:
Hello @sibirrer ! This PR is also ready!

sibirrer (approved):
Thank you very much @timedilatesme !

PR #471, opened 2026-09-03
More Accurate Quasar Host Matching

timedilatesme:
This PR aims to refactor the quasar host matching algorithm to make it more accurate and account for scatters in the scaling relations which were previosuly not present.

### DIFF
--- a/slsim/Sources/SourceCatalogues/QuasarCatalog/quasar_host_match.py
+++ b/slsim/Sources/SourceCatalogues/QuasarCatalog/quasar_host_match.py
@@ -1,7 +1,48 @@
+"""Match quasars drawn from a luminosity function to plausible host galaxies.
+
+The quasar catalog fixes the abundance and luminosity of the AGN; this
+module assigns each of them a host galaxy, a black hole mass and an
+Eddington ratio that are jointly consistent with
+
+* the M-sigma relation of Kormendy & Ho (2013) *including* its 0.29 dex
+  intrinsic scatter,
+* an Eddington ratio distribution, either the power law of Korytov et al.
+  (2019) or a lognormal calibrated on the broad-line quasars of Shen et al.
+  (2011),
+* the bolometric correction of Runnoe et al. (2012), anchored to the same
+  qsogen SED that generates the broad-band photometry.
+
+The assignment is a weighted draw from the joint posterior rather than a
+nearest-neighbour match, so the sampled Eddington ratios follow the input
+distribution and the resulting (M_BH, sigma) pairs scatter around the
+M-sigma relation instead of lying exactly on it.
+"""
+
+from functools import lru_cache
+import warnings
+
 import numpy as np
+import speclite.filters
+from astropy.cosmology import FlatLambdaCDM
 from astropy.table import hstack
 from scipy.interpolate import interp1d
 from tqdm import tqdm
+
+from slsim.Sources.SourceCatalogues.QuasarCatalog.qsogen import Quasar_sed, params_agile
+
+# Eddington luminosity per solar mass [erg s^-1], 4 pi G M_sun m_p c / sigma_T
+L_EDDINGTON_PER_MSUN = 1.2570e38
+
+# Runnoe et al. (2012), erratum table 1: L_iso = zeta * lambda L_lambda and
+# log L_iso = A + B log(lambda L_lambda), for lambda L_lambda in erg/s.
+RUNNOE12_BOLOMETRIC_CORRECTION = {
+    1450: {"zeta": 4.20, "A": 4.745, "B": 0.910},
+    3000: {"zeta": 5.18, "A": 1.852, "B": 0.975},
+    5100: {"zeta": 8.10, "A": 4.891, "B": 0.912},
+}
+
+# Reference monochromatic luminosity used to anchor the qsogen SED [log erg/s].
+_LOG_L3000_REFERENCE = 46.0
 
 
 def sample_eddington_rate(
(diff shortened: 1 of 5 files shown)


timedilatesme:
Hi @Henry-Best-01 ! This PR is ready for review now. Thank you so much for agreeing to review it ;). Also a [README](https://github.com/timedilatesme/slsim/tree/accurate-lagn-pop-scatters/slsim/Sources/SourceCatalogues/QuasarCatalog) file explains the new matching algorithm.

slsim/Sources/SourceCatalogues/QuasarCatalog/quasar_host_match.py
```diff
@@ -1,272 +1,304 @@
...
+    },
+}
+
+
+def log_black_hole_mass(galaxy_type, vel_disp=None, stellar_mass=None):
+    """Mean log10 black hole mass of each galaxy and the intrinsic scatter
+    about it, using the relation of its type (see
+    :data:`BLACK_HOLE_MASS_RELATIONS`).
+
+    :param galaxy_type: type of each galaxy, a key of the relations
```
Henry-Best-01, line 58: by "a key of the relations", do you mean the "red" or "blue" keys? I see this is possible to expand in the future, but maybe add an example at the end of this, something like:
'(e.g. "red", "blue", etc.)'

The code is clear, but someone who doesn't explicitly look at it could think galaxy type may relate to other keys such as "early", "spiral", and so on.

slsim/Sources/SourceCatalogues/QuasarCatalog/quasar_host_match.py
```diff
@@ -1,272 +1,304 @@
...
+    log_mass = np.empty(types.shape)
+    scatter = np.empty(types.shape)
+    for name, relation in BLACK_HOLE_MASS_RELATIONS.items():
+        selected = types == name
+        if not selected.any():
+            continue
+        values = properties[relation["property"]]
+        if values is None:
+            raise ValueError(
+                f"The '{name}' galaxies need a '{relation['property']}' column."
```
Henry-Best-01, line 83: Should this be "key/value pair" instead of column? It looks like relation is a dictionary, not a table.

slsim/Sources/SourceCatalogues/QuasarCatalog/quasar_host_match.py
```diff
@@ -1,272 +1,304 @@
...
+            )
+        values = np.asarray(values, dtype=float)[selected]
+        # a non-positive property gives -inf/nan here, which the matcher drops
```
Henry-Best-01, line 86: When you say the matcher drops a non-positive property, do you mean that it draws another value, or still returns a nan? Does this happen often, and do we need to worry about it? I guess if the value is so low that it's unmeasurable, it probably doesn't matter.

slsim/Sources/SourceCatalogues/QuasarCatalog/quasar_host_match.py
```diff
@@ -1,272 +1,304 @@
...
+    """Log10 bolometric luminosity [erg/s] for absolute magnitude M_i(z=2).
 
-    Notes
-    -----
-    The default parameters are based on observational constraints from
-    quasar luminosity function studies. (See Eq. 16 in Korytov et al. 2019, https://arxiv.org/abs/1907.06530)
+    :param m_i: absolute i-band magnitude M_i(z=2)
+    :param scatter: object-to-object scatter of the bolometric
+        correction [dex]
+    :param rng: random number generator used for the scatter
```
Henry-Best-01, line 101: I like the idea of passing in a generator object. Are there any worries about reproducibility though? If you pass an RNG with a certain state, can you easily and reliably get to that state without running every process prior to this calculation? I've never gotten and updated a generator's state and typically just set the seed.

Henry-Best-01: Alternatively, you could (also) pass in a seed, which people may be more accustomed to doing.

slsim/Sources/SourceCatalogues/QuasarCatalog/quasar_host_match.py
```diff
@@ -1,272 +1,304 @@
...
-
     def __init__(
         self,
         quasar_catalog,
         galaxy_catalog,
+        delta_z=0.001,
+        min_candidates=50,
+        max_delta_z=0.05,
+        bolometric_correction_scatter=0.1,
+        eddington_ratio_location=ERDF_LOCATION,
```
Henry-Best-01, line 131: Should this parameter be named "log_eddington_ratio_location", since it is expected as a log value?

slsim/Sources/SourceCatalogues/QuasarCatalog/quasar_host_match.py
```diff
@@ -1,272 +1,304 @@
...
+        :param max_delta_z: largest half-width the slice may be widened to
+        :param bolometric_correction_scatter: object-to-object scatter of the
+            bolometric correction [dex]
+        :param eddington_ratio_location: mean of log10(Eddington ratio)
+        :param eddington_ratio_width: standard deviation of log10(Eddington
+            ratio) [dex]
+        :param max_offset_sigma: a quasar is rejected if no candidate host can
+            produce it within this many combined standard deviations of the
+            mass relation and Eddington-ratio distribution
+        :param rng: random number generator, for reproducible catalogs
```
Henry-Best-01, line 158: Similar comment to previous generator comment, this is fine if there's no issue getting the generator to the expected state, otherwise an rng seed could be helpful.

slsim/Sources/SourceCatalogues/QuasarCatalog/quasar_host_match.py
```diff
@@ -1,272 +1,304 @@
...
+        :type rng: numpy.random.Generator or None
+        :param progress: whether to show a progress bar
```
Henry-Best-01, line 160: Does this take long for a catalog? Since it's on by default with True, maybe inverting the bool and naming it "quiet_mode" would be more intuitive than calling the parameter "progress".

Henry-Best-01: Another label could be "disable_tqdm", since it's linked to the tqdm "disable" parameter.

slsim/Sources/SourceCatalogues/QuasarCatalog/quasar_host_match.py
```diff
@@ -1,272 +1,304 @@
...
-            tqdm(
-                zip(self.quasar_catalog["z"], self.quasar_catalog["M_i"]),
-                total=len(self.quasar_catalog),
-                desc="Matching quasars with host galaxies",
-            )
+        self._validate()
+        galaxy_z, log_mass, scatter = self._prepare_galaxies()
+        quasar_z = np.asarray(self.quasar_catalog["z"], dtype=float)
+        log_l_bol = log_bolometric_luminosity(
+            self.quasar_catalog["M_i"], scatter=self.bc_scatter, rng=self.rng
```
Henry-Best-01, line 189: I am pretty sure that running the `match` method twice in a row will have different rng states. I think you need a way to pass a generator or a seed into `match` for reproducibility, since `QuasarHostMatch.rng` will update its state after doing any random process.

slsim/Sources/SourceCatalogues/QuasarCatalog/quasar_host_match.py
```diff
@@ -1,272 +1,304 @@
...
+
+        self._warn_about_rejections()
+        matched = np.array(matched).reshape(-1, 4)
+        quasar_index, galaxy_index = matched[:, :2].astype(int).T
+        quasars = self.quasar_catalog[quasar_index]
+        galaxies = self.galaxy_catalog[galaxy_index]
+        galaxies["black_hole_mass_exponent"] = matched[:, 2]
+        galaxies["eddington_ratio"] = 10 ** matched[:, 3]
+        galaxies["log_bolometric_luminosity"] = log_l_bol[quasar_index]
+        galaxies.remove_column("z")
```
Henry-Best-01, line 233: Is there a reason to remove the "z" column of galaxies? In the `_validate` method below, it looks like it is looking for this "z" column, so I must be misunderstanding something.

slsim/Sources/SourceCatalogues/QuasarCatalog/quasar_host_match.py
```diff
@@ -1,272 +1,304 @@
...
+        return hstack(
+            [quasars, galaxies],
             table_names=["quasar", "host_galaxy"],
             join_type="exact",
         )
 
-        return matched_catalog
+    def _validate(self):
```
Henry-Best-01, line 240: Needs a short docstring description. Something along the lines of "Validates all columns in current quasar_catalog and galaxy_catalog exist." is to the point.

slsim/Sources/SourceCatalogues/QuasarCatalog/quasar_host_match.py
```diff
@@ -1,272 +1,304 @@
...
+        ``min_candidates`` galaxies."""
+        delta_z = self.delta_z
+        while True:
+            start = np.searchsorted(galaxy_z, redshift - delta_z, side="left")
+            end = np.searchsorted(galaxy_z, redshift + delta_z, side="right")
+            if end - start >= self.min_candidates or delta_z >= self.max_delta_z:
+                return start, end
+            delta_z = min(delta_z * 2, self.max_delta_z)
+
+    def _warn_about_rejections(self):
```
Henry-Best-01, line 288: This needs a short docstring description.

Henry-Best-01 (changes requested):
Just some small adjustments and suggestions. Thanks for the PR!

PR #473, opened 2026-09-16
Add optional background subtraction for HST COSMOS and COSMOS-Web sources

FerroYx:
This PR adds optional constant-background subtraction to reduce background-related edge artifacts in catalog galaxy cutouts.

- HST COSMOS uses the matched galaxy's `NOISE_MEAN`, preserving background metadata during catalog preprocessing.
- COSMOS-Web estimates the background separately for each native band using a sigma-clipped median of outer pixels, excluding the central region and detected bright sources.
- Enable through `subtract_background=True` in `extended_source_kwargs`. The default remains `False`.
- Subtraction runs once per source before band mixing or color-gradient processing. Image dimensions and negative pixel values are preserved; no cropping, tapering, or denoising is applied.

This is an experimental mitigation rather than a complete solution to edge artifacts. Noise boundaries remain, and the COSMOS-Web estimator may oversubtract extended galaxy light when the cutout contains insufficient blank sky.

Validation includes synthetic-image checks and local HST catalog-to-image checks. Full population rendering validation is still pending.


FerroYx:
Under Construction......

PR #474, opened 2026-09-16
Bind Roman point-source paint to with_source

### LINKED ISSUE #360: point sources plotted when source=False
when creating images and only want to show the deflector, setting source=False removes the extended source but keeps the point sources

shaneraphel:
Closes #360

`simulate_image` already exposes independent `with_source` and `with_point_source` flags (both default True) and binds `point_source_add=with_point_source`. The earlier Roman dest collapsed those two channels onto `with_source`, so a host-off request also forced the point source off.

This change adds `with_point_source=True` on `simulate_roman_image` and binds `point_source_add=with_point_source`. Default True matches the LSST occupancy. It does not rewrite `simulate_image`.

| `with_source` | `with_point_source` | leftover `point_source_add` | dest `point_source_add` |
| --- | --- | --- | --- |
| False | True (default) | False | True |
| False | False | False | False |
| True | True | True | True |
| True | False | True | False |


ahuang314 (changes requested):
> It does not add a new `with_point_source` keyword

Maybe we should add it here too, for consistency

shaneraphel:
Added `with_point_source=True` on `simulate_roman_image` and bound `point_source_add=with_point_source`, matching `simulate_image`. Default True is the LSST occupancy: `with_source=False` no longer forces the point source off.

| `with_source` | `with_point_source` | leftover `point_source_add` | dest `point_source_add` |
| --- | --- | --- | --- |
| False | True (default) | False | True |
| False | False | False | False |
| True | True | True | True |
| True | False | True | False |

ahuang314 (approved):

sibirrer (approved):
Thank you @shaneraphel !

PR #475, opened 2026-09-16
Add u-band to Euclid-Roman-LSST YAML

Zhengjingyi0823:
Add u-band to euclid-roman-lsst-like.yaml


sibirrer:
I assigned @timedilatesme as a reviewer. There might be a reason we do not support as much the LSST-u band

sibirrer:
@timedilatesme I added you as a reviewer her, just to make sure things are ok with you

timedilatesme:
Hello @sibirrer ! Sounds good! Sorry was just writing the reply.

Skypy galaxies only allow u band fluxes being calculated for a redshift of `z < 4` but by default mostly we simulate galaxies down to z = 5 where skypy SEDs fails to give u band fluxes so that's why we haven't kept this as a default option.

There is a way to provide optional argument while making the skypy galaxy population [here](https://github.com/LSST-strong-lensing/slsim/blob/main/slsim/Pipelines/skypy_pipeline.py) the `z_max = 4` and filters = ['u', 'g', 'r', 'i', 'z', 'y'] [should work](https://github.com/LSST-strong-lensing/slsim/blob/38a426e02f8277a478ffd35910166d6acfc9c55f/slsim/Pipelines/skypy_pipeline.py#L34-L36).

So I think we should not hardcode the u band in there.

@Zhengjingyi0823 you can provide the optional arguments during the instantiation of skypypipeline so that we don't modify the default config.

sibirrer:
ah right @timedilatesme. Can we perhaps make the skypy pipeline result overwrite Nan's with magnitude =1000 or so, just to make sure it can be processed? I don't expect u-band flux from high-z lenses anyways

timedilatesme:
Sounds great @sibirrer ! Will make these changes and push the PR.

